PR #16, opened 2018-06-29
Adding `id` property to `Page` and `Button` components

DanielSchiavini:


ChrisTerBeke:
Wouldn't it be nicer to give pages a unique ID instead?

PR #18, opened 2018-07-13
[WIP] Fixes for parcel setup

ChrisTerBeke:
* Fixes classnames import (typescript issue)
* Makes font loading URLs from SASS relative
* Remove hot-reloading (and createApp method)
* Add new i18n utility that allows loading the po files from TypeScript
* Cleanup dependencies (remove babel, webpack, etc)


src/views/page_not_found.tsx
```diff
@@ -0,0 +1,32 @@
...
+    location: Location
+}
+
+export interface PageNotFoundViewState {
+}
+
+/**
+ * The page not found view.
+ */
+export default class PageNotFoundView extends React.Component<PageNotFoundViewProps, PageNotFoundViewState> {
```
ChrisTerBeke: Maybe we should look into showing an error here (response_error.tsx) as we want users to report this because it shouldn't happen in the first place.

src/components/header.tsx
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2018 Ultimaker B.V.
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import CuraLogo from './cura_logo';
+import CuraLogo from './icons/cura_logo';
```
ChrisTerBeke: We should also add the Ultimaker logo here.

ChrisTerBeke:
@woosie10 the tests started failing.

PR #20, opened 2018-07-17
Create slide in panel

woosie10:


ChrisTerBeke:
@woosie10 Can your rebase this on the parcel branch (as that's heavily used in the new app already)?

src/components/slide_in_panel.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+      <Motion style={{ x: spring(isOpen ? 0 : 100, motion) }}>
+        {({ x }) =>
+          <div className="slide-in-panel__container" style={{ transform: `translate3d(${x}%,0,0)`, width: panelWidth }}>
+            <div className="slide-in-panel__header">
+              {headerTitle}
+            </div>
+            <div className="slide-in-panel__content">
+              {children}
+            </div>
```
ChrisTerBeke: Looking at the design, it might also need an optional footer (for the buttons)?

src/components/slide_in_panel.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  render(): JSX.Element {
+    const { headerTitle, isOpen, panelWidth, children } = this.props;
+
+    const motion = { stiffness: 450, damping: 50 };
+    const classes = classNames('slide-in-panel', { isOpen });
+
+    return <div className={classes} onClick={(e) => this._handleOverlayClick(e)}>
+
+      <Motion style={{ x: spring(isOpen ? 0 : 100, motion) }}>
```
ChrisTerBeke: Not sure how the motion is calculated, but it seems a bit jittery. Is it pure js or css? Css would be better as that's hardware accelerated.

src/components/slide_in_panel.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export class SlideInPanel extends React.Component<SlideInPanelProps, {}> {
+
+  _handleOverlayClick(e: React.MouseEvent<HTMLDivElement>): void {
+    const { onOverlayClickHandler } = this.props;
+
+    e.stopPropagation();
+
+    if (onOverlayClickHandler) {
+      onOverlayClickHandler();
```
ChrisTerBeke: Maybe default this to closing the slide-out?

src/components/slide_in_panel.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export interface SlideInPanelProps {
+  /** Text to be displayed in the panel header */
+  headerTitle: string;
+  /** The panel will be open when true */
+  isOpen: boolean;
+  /** Called when the background overlay is clicked */
+  onOverlayClickHandler?: Function;
+  /** Width of panel. Include unit */
+  panelWidth?: string;
```
ChrisTerBeke: The name `width` would suffice I think as it's already in the context of the slide-in panel.

src/components/slide_in_panel.tsx
woosie10: This does use js. Is the Slide out container component also jittery for you?

ChrisTerBeke: Yup.

src/components/response_error.tsx
```diff
@@ -0,0 +1,165 @@
...
+        at least one number and both lower and uppercase letters and special characters.",
+    "badPasswordRepeat": "Repeated password did not match original password.",
+    "dataUriImageInvalid": "Profile image was not valid. Please use a JPEG or PNG image."
+
+}
+
+/**
+ * Map of all form field names that should be translated.
+ */
+export const FIELD_NAME_TO_I18NC = {
```
DanielSchiavini: this should not be in the `response_error` file. We should have a separate file with translations

src/components/slide_in_panel.tsx
```diff
@@ -0,0 +1,61 @@
...
+  /** Text to be displayed in the panel header */
+  headerTitle: string;
+  /** The panel will be open when true */
+  isOpen: boolean;
+  /** Called when the background overlay is clicked */
+  onOverlayClickHandler?: Function;
+  /** Width of panel. Include unit */
+  width?: string;
+  /** Show footer */
+  includeFooter?: boolean;
```
DanielSchiavini: should we include children here to explain how they are interpreted?

woosie10: I don't think you can do that

PR #31, opened 2018-08-07
Translatable errors

ChrisTerBeke:


src/components/response_error.tsx
```diff
@@ -8,63 +8,30 @@ import Tile from './tile';
...
-
-    // user data
-    "username": I18n.translate("form field username", "Username"),
-    "email": I18n.translate("form field email", "Email Address"),
-    "name": I18n.translate("form field name", "Full Name"),
-    "password": I18n.translate("form field password", "Password"),
-    "password_repeat": I18n.translate("form field password_repeat", "Repeat Password"),
-
-}
+export type ErrorMessageTemplates = {[key: string]: string};
```
DanielSchiavini: add something like:
// a mapping of error codes to translated error messages which may include the error parameters

and
// a mapping of field names to their translated descriptions

src/components/response_error.tsx
```diff
@@ -8,63 +8,30 @@ import Tile from './tile';
...
+            if (key === "field_name" && fieldTranslations[value]) {
+                value = fieldTranslations[value]
             }
             params[key] = value
         });
     }
-    return I18n.format("error code", ERROR_CODE_TO_I18NC_TEMPLATE[error.code]
-        || ERROR_CODE_TO_I18NC_TEMPLATE["genericError"], params)
+    const errorMessage = errorMessageTemplates[error.code] || I18n.translate('error message default',
+        "There was a problem on the server, please try again. (Code: %{code})")
```
DanielSchiavini: Can we maybe include the title here? Something like `"There was an issue on the server: {title} (code {code})"`

src/components/response_error.tsx
ChrisTerBeke: Done.

src/components/response_error.tsx
ChrisTerBeke: Done.

PR #37, opened 2018-08-09
Changes for cura connect integration

woosie10:


src/components/slide_out_container.tsx
```diff
@@ -50,6 +48,10 @@ export class SlideOutContainer extends React.Component<SlideOutContainerProps, S
     </div>
 
   }
-}
+
+SlideOutContainer.defaultProps = {
+  showHeaderPanelArrow: true,
+  headerTextAlignment: 'left'
+};
```
ChrisTerBeke: missing SlideOutContainer.displayName = 'SlideOutContainer'

ChrisTerBeke: Also for many other components I see (this makes the snapshots know what the names of components are).

src/components/slide_in_panel.tsx
```diff
@@ -24,6 +24,8 @@ export interface SlideInPanelProps {
 
 export class SlideInPanel extends React.Component<SlideInPanelProps, {}> {
 
+  public static displayName = 'SlideInPanel';
```
ChrisTerBeke: For classes it's not needed as they are already named by their class.

woosie10: Ahhhhhh, I'll remove them

PR #41, opened 2018-08-14
add flag to display a ContextMenu in a panel

woosie10:


src/components/context_menu.tsx
```diff
@@ -13,6 +13,8 @@ export interface ContextMenuProps {
   menuOffsetDirection?: MenuOffsetDirection;
   /** Direction to position the menu: 'north' | 'south' */
   menuDirection?: MenuDirection;
+  /** Whether the context menu is position in a panel, such as a header or footer */
+  menuPanelPosition?: boolean;
```
ChrisTerBeke: Can't we set the correct CSS rules automatically by checking if it's a child element inside a panel component?

sedwards2009: This doesn't sound like a `boolean` property.

PR #45, opened 2018-08-20
Make inputs controlled

woosie10:
It give us more flexibility in using the input components they should be changed to controlled components


src/components/tags_selector.tsx
```diff
@@ -53,20 +51,19 @@ export class TagsSelector extends React.Component<TagsSelectorProps, TagsSelecto
 
   static getDerivedStateFromProps(props: TagsSelectorProps, state: TagsSelectorState): TagsSelectorState {
     let suggestions = []
-    let defaultTags = []
+    let tags = []
 
-    if (props.suggestions && !state.defaultSet) {
+    if (props.suggestions && TagsSelector._convertStringsToTags(props.suggestions) !== state.suggestions) {
```
DanielSchiavini: We are calling _convertStringsToTags here more often than we should need to

src/components/popup.tsx
```diff
@@ -72,6 +72,14 @@ export class Popup extends React.Component<PopupProps, PopupState> {
     this._secondaryBtnHandler = this._secondaryBtnHandler.bind(this);
   }
 
+  componentDidMount(): void {
+    const { inputDefaultValue } = this.props;
+
+    if (inputDefaultValue){
+      this.setState({inputValue: inputDefaultValue.toString()})
```
DanielSchiavini: Why are we copying the props to state?

src/components/form.tsx
```diff
@@ -2,7 +2,9 @@ import * as React from 'react';
 import { Link } from 'react-router-dom';
 
 import { default as Button, ButtonStyle } from './button';
-import InputField from './input_field'
+
+
+export type FormValidationResponse = { [key: string]: any[] | string }
```
DanielSchiavini: maybe it's nice to explain why it can be any[] cause I also asked myself the same thing today :stuck_out_tongue_closed_eyes: 

We also might just use `any`

src/components/form.tsx
```diff
@@ -20,7 +22,7 @@ export interface FormProps {
 	/** A url to link to instead of calling secondaryBtnHandler */
 	secondaryBtnLink?: string;
 	/** The form validation state and validation error messages */
-	formValidation?: FormValidationResponse;
+	formValidationErrors?: FormValidationResponse;
```
DanielSchiavini: We are already in the form so I'd name this `validationErrors`

src/components/input_field.tsx
```diff
@@ -22,10 +22,8 @@ export interface InputFieldProps {
   className?: string;
   /** Input field will be centered if true. Useful for type image or checkbox */
   centerInputField?: boolean;
-  /** Input field will be displayed in the error state when true */
-  validationError?: boolean;
   /** Message to show for the validation error */
-  validationErrorMsg?: string;
+  validationError?: string;
```
DanielSchiavini: `| any[]`

src/components/input_field_validation.tsx
```diff
@@ -5,7 +5,7 @@ export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg';
 
 export interface InputFieldValidationProps {
   /** Message to show for the validation error */
-  validationErrorMsg: string;
+  validationError: string;
```
DanielSchiavini: `| any[]`

src/components/popup.tsx
woosie10: To set the default value of the input field. This is only happening on componentDidMount. I will add a comment

src/components/tags_selector.tsx
```diff
@@ -50,20 +50,23 @@ export class TagsSelector extends React.Component<TagsSelectorProps, TagsSelecto
   }
 
   static getDerivedStateFromProps(props: TagsSelectorProps, state: TagsSelectorState): TagsSelectorState {
-    let suggestions = []
-    let tags = []
+    let newSuggestions: Tag[] = null;
+    let newTags: Tag[] = null;
```
DanielSchiavini: shouldnt this be state.tags by default?

ChrisTerBeke:
The folder 'public' should be removed and ignored.

PR #55, opened 2018-08-31
Tidy and configure

woosie10:
- change 2 space indenting to 4
- change sass to scss (cmd: `sass-convert -R ./ -F sass -T scss && find . -type f -name '*.sass' -delete`)
- add some more comments


ChrisTerBeke:
We'll also have to update the apps to use scss.

PR #56, opened 2018-09-03
More frontend tests

woosie10:


src/components/__tests__/image_upload.test.tsx
```diff
@@ -1,26 +1,42 @@
...
-
-  it('should render', () => {
-    expect(wrapper).toMatchSnapshot();
-    expect(props.onFileSelection).not.toHaveBeenCalled();
-  });
+    let props;
+    let wrapper;
+
+    beforeEach(() => {
+        props = {
```
ImreBognarUltimaker: there is an issue with react-dropzone that is generating an error:
https://github.com/react-dropzone/react-dropzone/issues/554

src/components/__tests__/context_menu.test.tsx
```diff
@@ -1,25 +1,74 @@
...
+    });
+
+    it('should render', () => {
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should render to right', () => {
+        wrapper.setProps({ menuOffsetDirection: 'right' });
+        expect(wrapper.find('.menu').prop('style')).toHaveProperty('left', -30);
+        expect(wrapper.find('.menu').prop('style')).toHaveProperty('right', undefined);
```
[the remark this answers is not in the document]
ImreBognarUltimaker: As we discussed, this would not be a good idea in this case.

src/components/__tests__/context_menu.test.tsx
```diff
@@ -1,25 +1,74 @@
...
+        expect(wrapper.find('.visible')).toHaveLength(1);
+        wrapper.find('.container').simulate('click', mockClickEvent);
+        expect(wrapper.find('.visible').exists()).toEqual(false);
+    });
 
+    it('should not propagate click', () => {
+        wrapper.simulate('click', mockClickEvent);
+        expect(mockClickEvent.stopPropagation).toBeCalled();
+    });
 });
```
ImreBognarUltimaker: We agreed on the following:
-Use .toHaveLength() to check for element existence, because it also checks if it exists only once.
-In the future, use .hasClass() where applicable. It's not worth the effort to rewrite anything for this now.
-Use .prop('disabled') to check for disabled status, instead of .find([disabled])
-Use toBe() when not comparing objects.
-Use .prop('key') instead of .props().key.

src/components/__tests__/context_menu.test.tsx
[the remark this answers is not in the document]
ImreBognarUltimaker: There is a recommendation to use .toBeUndefined() when checking for undefined.
https://jestjs.io/docs/en/expect#tobeundefined

src/components/__tests__/form.test.tsx
```diff
@@ -1,26 +1,85 @@
...
+
+    it('should render', () => {
+        expect(wrapper).toMatchSnapshot();
+        expect(props.onSubmitHandler).not.toHaveBeenCalled();
+    });
+
+    it('should call onSubmitHandler when form is submitted', () => {
+        wrapper.simulate('submit', mockClickEvent);
+        expect(props.onSubmitHandler).toHaveBeenCalled();
+        expect(wrapper.state().submitted).toEqual(true);
```
ImreBognarUltimaker: I propose to use state() as we agreed in the case of props().
state('key') instead of state().key

src/components/__tests__/slide_out_container.test.tsx
```diff
@@ -1,27 +1,51 @@
 // Copyright (c) 2018 Ultimaker B.V.
-import 'jest';
-import 'jsdom-global/register';
 import * as React from 'react';
 import { shallow } from 'enzyme';
+import { UnmountClosed } from 'react-collapse';
```
ImreBognarUltimaker: This import should be a few lines below in the component list.

src/components/__tests__/select_list.test.tsx
```diff
@@ -1,32 +1,70 @@
...
+    it('should not render a label when value is null', () => {
+        wrapper.setProps({ value: null });
+        expect(wrapper.find('.text').text()).toEqual('');
+    });
+
+    it('should hide menu when the label is clicked', () => {
+        wrapper.find('.label').simulate('click', mockClickEvent);
+        expect(wrapper.find('.visible')).toHaveLength(1);
+        wrapper.find('.label').simulate('click', mockClickEvent);
+        expect(wrapper.find('.visible').exists()).toEqual(false);
```
ImreBognarUltimaker: What is the reason here for having 2 different patterns to check for the existence of an element with this class?

src/components/__tests__/image_upload.test.tsx
woosie10: updated

src/components/__tests__/context_menu.test.tsx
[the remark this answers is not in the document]
woosie10: updated

src/components/__tests__/context_menu.test.tsx
woosie10: updated

src/components/__tests__/slide_out_container.test.tsx
woosie10: I normally put external imports at the top, then our components under `// components`. react-collapse is an external import.

src/components/__tests__/select_list.test.tsx
woosie10: It's not, the first checks that it's open, the the second checks that it's closed

ImreBognarUltimaker:
Everything looks good.

PR #58, opened 2018-09-14
Component improvements

woosie10:
This PR makes the following changes
* Use the Form component within the Popup component
  * this contains some breaking changes. Prompt popups should now use the component PopupPrompt and the `type` should be removed from all Popup components
* Fix the height of the alert pill button, so that it is the same as the pill buttons
* Allow the InputField component to show the preLabelElement even when now label is passed
* Fix the checkbox label alignment so that it is centred vertically
* Prevent the use of spaces in tags in the TagsSelector component
* Add a Textarea component that can auto grow
* Add the ability to add a clickable area around the profile picture of the UserAccountMenu component
* Show validation messages when InputField is type 'children'
* Improves the organisation of colour variables and consolidates some colours
* Add all Cura connect icons
* Show all colors in storybook
* Add `size` and `color` props to all icons


ChrisTerBeke:
Anything special in the dev portal we need to do to adopt these (after they're in a release)?

woosie10:
> Anything special in the dev portal we need to do to adopt these (after they're in a release)?

I've added

src/components/button.tsx
```diff
@@ -46,19 +46,19 @@ export const Button: React.StatelessComponent<ButtonProps> =
             }
         }
 
-        if (type === 'link' && isLinkExternal) {
+        if (type === 'link' && isLinkInternal) {
```
ChrisTerBeke: Not a big fan of this if/else if/else thing. Maybe use a switch, or better even map the types to a render function.

src/components/form.tsx
```diff
@@ -54,9 +58,21 @@ export class Form extends React.Component<FormProps, FormState> {
         this.props.onSubmitHandler();
     }
 
+    _isPrimaryBtnDisabled() {
```
ChrisTerBeke: Missing private

src/components/form.tsx
```diff
@@ -69,15 +85,15 @@ export class Form extends React.Component<FormProps, FormState> {
             <div className="form__item">
                 {React.cloneElement(child, errors && {
                     validationError: errors,
-                    submitted
+                    submitted: child.props.submitted || submitted
```
ChrisTerBeke: Will this crash if submitted is not a property on the child?

src/components/icons/icon_wrapper.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+export type Size = 'sm' | 'md' | 'lg';
```
ChrisTerBeke: Maybe call these IconSize and IconColor to prevent polluting the namespace when importing these (too generic).

src/components/input_field_input.tsx
```diff
@@ -102,66 +104,54 @@ export class InputFieldInput extends React.Component<InputFieldInputProps, {}> {
...
             value, imageSize, staticField, imageShape, tagSuggestions, focusOnLoad,
-            showValidationError, onChangeHandler, children } = this.props;
+            showValidationError, onChangeHandler, textareaAutoGrow, children } = this.props;
 
         const classes = classNames('input', { 'error': showValidationError })
 
         if (type === 'children') {
             return children;
         }
         if (type === 'textarea') {
```
ChrisTerBeke: Same as above, either a switch but preferably a type -> render method map.

src/components/popup.tsx
```diff
@@ -42,151 +32,78 @@ export interface PopupProps {
...
-        const validationErrorMsg = this.props.validationHandler(inputValue);
-
-        if (!validationErrorMsg) {
-            return true;
-        }
-        else {
-            this.setState({ validationErrorMsg: validationErrorMsg });
-            return false;
-        }
+    _primaryBtnHandler(): void {
```
ChrisTerBeke: missing protected

src/components/popup_prompt.tsx
```diff
@@ -0,0 +1,128 @@
...
+    componentDidMount(): void {
+        const { inputDefaultValue } = this.props;
+
+        if (inputDefaultValue) {
+            // set the initial value of the prompt input field
+            this.setState({ inputValue: inputDefaultValue.toString() })
+        }
+    }
+
+    _onChangeHandler(id: string, value: InputFieldValue): void {
```
ChrisTerBeke: missing private (multiple in this file).

src/components/textarea.tsx
```diff
@@ -0,0 +1,64 @@
...
+    constructor(props) {
+        super(props);
+        this.inputRef = React.createRef();
+    }
+
+    componentDidMount(): void {
+        this._focusOnPromptInput();
+    }
+
+    _focusOnPromptInput(): void {
```
ChrisTerBeke: missing private

src/components/textarea.tsx
```diff
@@ -0,0 +1,64 @@
...
+    autoGrow?: boolean;
+    /** html placeholder text */
+    placeholder?: string;
+    /** Whether the textarea should be auto-focused */
+    autofocus?: boolean;
+}
+
+export class Textarea extends React.Component<TextareaProps, {}> {
+
+    private inputRef;
```
ChrisTerBeke: should start with underscore as it's private.

src/components/textarea.tsx
```diff
@@ -0,0 +1,64 @@
...
+    /** Called when the checkbox is clicked */
+    onChangeHandler: (value: string) => void;
+    /** When true the error state will be displayed */
+    error?: boolean;
+    /** Whether the textarea should grow horizontally with user input */
+    autoGrow?: boolean;
+    /** html placeholder text */
+    placeholder?: string;
+    /** Whether the textarea should be auto-focused */
+    autofocus?: boolean;
```
ChrisTerBeke: autoGrow is camelcased but autofocus is not?

src/components/form.tsx
woosie10: No, `child.props.submitted` would be undefined, and so it would use `submitted`, which by default is false

src/components/textarea.tsx
woosie10: Autofocus is one word

src/components/popup.tsx
woosie10: Why should this one be protected, but the others private? Shouldn't it also be private?

ChrisTerBeke:
@woosie10 What I mean with mapping the button (and other components) type to render function was something like this:

```typescript
private _typeToRenderMap = {
   "link": this._renderLink,
    "otherType": this._renderOtherType
}

...

const { type } = this.props
return this._typeToRenderMap[type]()
```

That omits the usage of an if/else tree that is hard to read.

woosie10:
> @woosie10 What I mean with mapping the button (and other components) type to render function was something like this:
> 
> ```ts
> private _typeToRenderMap = {
>    "link": this._renderLink,
>     "otherType": this._renderOtherType
> }
> 
> ...
> 
> const { type } = this.props
> return this._typeToRenderMap[type]()
> ```
> 
> That omits the usage of an if/else tree that is hard to read.

I see that would be useful for InputFieldInput, but not really for the Button, as that if is not just on the type, but also on the return of _isLinkInternal()

PR #59, opened 2018-09-24
Implementing ratings component

DanielSchiavini:


src/components/__tests__/ratings.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+import HighlightIcon from '../icons/highlight_icon'
+import ApprovedIcon from '../icons/approved_icon'
+
+describe('The Ratings component', () => {
+    it('should render 4.3 of 5 highlight icons', () => {
+        let wrapper = shallow(<Ratings max={5} rating={4.3} Icon={HighlightIcon} />);
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should render a third of 10 approved icons', () => {
```
ChrisTerBeke: nitpicking: 3 out of 10. a third would be 3.3333333333...

DanielSchiavini: The point is exactly to test 3.33333 and see it works correctly. We pass `10 / 3`

DanielSchiavini:
@woosie10 I thought it would be nice to add this component to the shared components. Chris noticed a small border around the yellow components, do you know if it's possible to remove it?

src/components/ratings.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * The ratings component shows the ratings score of an object.
+ * @param max - The maximum rating allowed (integer).
+ * @param rating - The actual rating of the object.
+ * @param Icon - The icon class to use. It should be wrapped by `IconWrapper`.
+ * @param foregroundColor - The color of the foreground icons. Defaults to 'orange'.
+ * @param backgroundColor - The color of the background icons. Defaults to 'grey'.
+ * @constructor
+ */
+export const Ratings: React.StatelessComponent<RatingsProps> = (
+    { max, rating, Icon, foregroundColor = "orange", backgroundColor = "grey" }
```
woosie10: for the other components we are using defaultProps to set the default. The benefit to this is that they then show in the info on storybook

src/stories/other.stories.tsx
```diff
@@ -21,6 +25,24 @@ stories.addDecorator(withKnobs)
...
+
+        <Ratings max={10} rating={4.3} Icon={HighlightIcon} />
+
+        <Ratings max={10} rating={0} Icon={HighlightIcon} />
+
+        <Ratings max={5} rating={3.5} Icon={ApprovedIcon} />
+
+        <Ratings max={5} rating={5 / 3} Icon={ApprovedIcon} />
+
+        <Ratings max={5} rating={null} Icon={ApprovedIcon} />
```
woosie10: Instead of having these multiple times, it would be  better to use a knob, so that the user can set the props themselves through storybook

PR #61, opened 2018-09-28
Refactoring input fields

DanielSchiavini:
I'm not happy with the fact that we are exporting two components per file, static and non-static. Some files look very convoluted and duplicated in cases like checkbox and tags.

The good thing is that this is backwards compatible and we can improve it further later on.


ChrisTerBeke:
In general I like this change. It's a bit more duplicated code, but gives the following benefits:

* Knowing exactly which props apply to which input field type
* Having the field types stateless, and having the state logic in a special wrapper component
* Having business logic separated per field type (e.g. no date field logic polluting a text area)

@ImreBognarUltimaker and @woosie10 can you have a look as well?

DanielSchiavini:
It also makes it easier to add new input types without polluting the existing components. We could now easily replace the 'children' type with actual components

DanielSchiavini:
Check out #63

src/components/input_field.tsx
```diff
@@ -42,152 +47,41 @@ export interface InputFieldProps {
...
 }
 
-export class InputField extends React.Component<InputFieldProps, InputFieldState> {
-
-    public static defaultProps: Partial<InputFieldProps> = {
+/**
+ * Class that adds an input wrapper around a ImageUpload component.
+ * TODO: merge the two components
+ */
+export class InputField extends React.Component<OldInputFieldProps, {}> {
```
DanielSchiavini: This component should be completely backwards compatible, but using the specific fields would provide better typing and flexibility

src/components/file_upload.tsx
```diff
@@ -49,13 +55,18 @@ export class FileUpload extends React.Component<FileUploadProps, {}> {
...
             />
             <div className="layout layout--gutter-sm">
                 <div className="layout__item u-fill file-upload__input-container">
                     <label className={inputClasses} htmlFor={id}>{selectedFileName}</label>
                 </div>
                 <div className="layout__item u-fit">
-                    <label className='btn btn--primary' htmlFor={id}>{I18n.translate('file upload button', 'Choose file')}</label>
+                    <label className='btn btn--primary' htmlFor={id}>
+                        {I18n.translate('file upload button', 'Choose file')}
+                        {/* TODO: Check with Alan which classes we need to have the spinner keep the button size */}
```
ChrisTerBeke: I saw some issues here where the text and spinner were both visible at the same time.

src/components/input_field.tsx
```diff
@@ -42,152 +50,42 @@ export interface InputFieldProps {
...
-        if (this._showValidationError() && validationError) {
-            return <InputFieldValidation
-                validationError={validationError}
-                labelLayoutWidth={labelLayoutWidth}
-                labelWidthBreakpoint={labelWidthBreakpoint}
-                required={required}
-            />
+    render() {
+        const {type, value, children, ...wrapperProps} = this.props;
+        if (type === 'text' || type === 'password' || type === 'email' || type === 'url') {
```
ChrisTerBeke: I wouldn't be against also just putting these types in the map above.

src/components/input_fields/__tests__/input_field_validation.test.tsx
```diff
@@ -6,35 +6,35 @@ import { shallow } from 'enzyme';
 import InputFieldValidation from '../input_field_validation';
 
 describe('The InputFieldValidation component', () => {
-    let props;
-    let wrapper;
+    let props
```
ChrisTerBeke: This file is missing all ; now as only one.

src/components/input_fields/checkbox_field.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * The checkbox field keeps track of whether it has been touched.
+ */
+export interface CheckboxFieldState {
+    /** Indicates if the field has been touched (changed) or not from the default value. */
+    touched: boolean;
+}
+
+/**
+ * Class that adds an input wrapper around a Checkbox component.
+ * TODO: merge the two components
```
ChrisTerBeke: Which two components?

src/components/file_upload.tsx
DanielSchiavini: @woosie10 indeed can you help on this one maybe? the button gets a different size when we hide the text.

src/components/input_field.tsx
DanielSchiavini: We'd have to pass type to each input field but that's fine

src/components/input_fields/__tests__/input_field_validation.test.tsx
DanielSchiavini: pycharm :cry:

src/components/input_fields/checkbox_field.tsx
DanielSchiavini: checkbox_field and checkbox

ChrisTerBeke:
@woosie10 when can you review this? It's blocking us from finishing a bug before the end of the sprint.

src/components/input_fields/image_upload_field.tsx
```diff
@@ -0,0 +1,83 @@
...
+import ImageUpload, {ImageFile} from '../image_upload';
+import {Image, ImageShape} from '../image';
+
+
+/**
+ * The image upload field provides these props in addition to those supported by all input fields.
+ */
+export interface ImageUploadFieldProps extends InputFieldProps {
+    /** Size of the image for type image. Include size unit */
+    imageSize: string;
```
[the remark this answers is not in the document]
DanielSchiavini: Before they couldn't be mandatory because the props were applicable to all input types. Now we have a specific type for images we should always pass an image size. Although I understand HTML will default it to the actual size of the image so this can be optional

src/components/input_fields/file_upload_field.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * The checkbox field provides these props in addition to those supported by all input fields.
+ * Note that this field does not have a value prop.
+ */
+export interface FileUploadFieldProps extends InputFieldProps {
+    /** Called when a file is selected */
+    onChangeHandler?: (id: string, target: HTMLInputElement) => void;
+    /** Called when a file is read. The promise should be fulfilled when the file is ready uploading. */
+    onReadHandler?: (id: string, value: string) => void;
```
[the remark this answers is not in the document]
DanielSchiavini: then the component cannot be backwards compatible. Also I can imagine reading the file in other ways than the one we implemented in the field, that's why I left the two possible callbacks

src/components/input_fields/wrapped_input_field.tsx
```diff
@@ -0,0 +1,94 @@
...
+export interface WrappedInputFieldState {
+    /** Indicates if the field has been touched (changed) or not from the default value. */
+    touched: boolean;
+}
+
+/**
+ * Class that adds an input wrapper around a HTML input component.
+ * The children passed are used only in the static view!
+ */
+class WrappedInputField extends React.Component<WrappedInputFieldProps, WrappedInputFieldState> {
```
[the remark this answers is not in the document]
DanielSchiavini: sure sounds like a good idea, I had problems to come up with a better name

src/components/input_fields/file_upload_field.tsx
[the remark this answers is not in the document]
woosie10: ok that sounds fine

src/components/input_fields/checkbox_field.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    render() {
+        const {value, className, labelLayoutWidth, ...wrapperProps} = this.props;
+        const {id, staticField} = wrapperProps;
+        const {touched} = this.state;
+        return (
+            <InputFieldWrapper
+                touched={touched}
+                className={classNames(className, "input-field--checkbox")}
+                labelLayoutWidth="fill"
```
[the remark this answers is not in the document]
DanielSchiavini: To handle this old if:
https://github.com/Ultimaker/react-web-components/blob/e778c4a/src/components/input_field_input.tsx#L86

src/components/ratings.tsx
```diff
@@ -9,7 +9,9 @@ export interface RatingsProps {
     Icon: React.StatelessComponent<IconWrapperProps>;
```
woosie10: this type needs to be extended to add `borderColor`, as I currently get an error

PR #71, opened 2018-10-09
Adding 'accept' to FileUpload, giving opacity=0 instead of display=none

DanielSchiavini:
This PR makes the 'accept' prop available for the file input field.

Also it changes the styling: With opacity=0 and position=absolute the file input field should not be displayed, but is still be available by using the tab character and clicking on the labels.

### DIFF
--- a/src/stylesheets/_forms.scss
+++ b/src/stylesheets/_forms.scss
@@ -299,6 +299,7 @@
     input {
         position: absolute;
         opacity: 0;
+        left: -9999px;
     }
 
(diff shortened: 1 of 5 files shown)


src/stylesheets/_forms.scss
```diff
@@ -297,7 +297,8 @@
 
 .file-upload {
     input {
-        display: none;
+        position: absolute;
```
woosie10: I would also add `left: -9999px` to move it off screen

PR #72, opened 2018-10-10
Cropping images

DanielSchiavini:
This PR adds the possibility of cropping images to the image upload component.


src/components/input_fields/image_upload_field.tsx
```diff
@@ -17,16 +17,19 @@ export interface ImageUploadFieldProps extends InputFieldProps {
     imageShape?: ImageShape;
 
     /** Called when the field changes */
-    onChangeHandler: (id: string, value: ImageFile) => any;
+    onChangeHandler?: (id: string, value: ImageFile) => any;
 
     /** Called when an image file is read */
-    onReadHandler: (id: string, file: ImageFile, contents: string) => any;
+    onReadHandler?: (id: string, dataURL: string) => any;
```
DanielSchiavini: This is a breaking change. We could keep the file in the state but the clients using this can also do it if necessary. The file.preview is not necessary, you can also use the dataURL given.

ChrisTerBeke: I think it's fine, we use this only in cloud front-ends at the moment (in 3? places).

src/components/image_cropper.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * The avatar editor requires an integer for size, so we assume the size is either px or rem.
+     * This may need to be replaced by a more complex logic in case we need to support relative sizes.
+     * @param size - The size of the image, including the unit.
+     * @param borderSize - The size of the border, in pixels.
+     * @return The size in pixels.
+     */
+    private _calculateSize(size: string, borderSize: number): number {
+        const sizeScale = size.endsWith("rem") ? 10 : 1;
```
ChrisTerBeke: Nitpick: single quotes. We should add a linter indeed to automate this checking.

ChrisTerBeke:
```suggestion
        const sizeScale = size.endsWith('rem') ? 10 : 1;
```

src/components/image_upload.tsx
```diff
@@ -30,71 +38,101 @@ export interface ImageUploadProps {
...
             const reader = new FileReader();
-            reader.onload = () => onFileRead(file, reader.result as string);
-            reader.onerror = console.error; // TODO
+            reader.onload = () => onFileRead(reader.result as string);
+            reader.onerror = console.error;
             reader.readAsDataURL(file);
         }
+
+        if (allowCropping) {
+            return this.setState({ cropURL: file.preview });
```
ChrisTerBeke: This method has a return type `void`, so either that type or this line should be changed.

src/components/range_slider.tsx
```diff
@@ -0,0 +1,48 @@
...
+        type="range"
+        className={classNames("range-slider", className)}
+        onChange={e => onChange(parseFloat(e.target.value))}
+        min={min}
+        max={max}
+        step={step}
+        defaultValue={value && value.toString()}
+    />;
+
+RangeSlider.displayName = "RangeSlider";
```
ChrisTerBeke: Nitpick: double quotes.

```suggestion
RangeSlider.displayName = 'RangeSlider';
```

ChrisTerBeke:
For me it works fine. Only UX thingy I have is that the image is smaller when cropping, so it's hard to get a good idea of the final result. Maybe we can have the 'grey area' be a bit bigger and leave the image size the same in the form.

src/components/image_cropper.tsx
```diff
@@ -0,0 +1,147 @@
...
+export class ImageCropper extends React.Component<ImageCropperProps, ImageCropperState> {
+    /* The default props */
+    static defaultProps: ImageCropperProps = {
+        shape: 'square',
+        size: '200px',
+        minScale: 1,
+        maxScale: 2.5,
+        scaleStep: 0.1,
+        borderSize: 25,
+        imageURL: '/images/Ultimaker-Resources.jpg', // TODO: Import file instead.
```
woosie10: Why do we need a default for this?

src/components/image_cropper.tsx
```diff
@@ -135,9 +134,7 @@ export class ImageCropper extends React.Component<ImageCropperProps, ImageCroppe
                     value={scale}
                 />
                 {onCropCancel &&
-                    <Button onClickHandler={onCropCancel} style="quiet" shape="pill" className="remove-image">
-                        <CrossIcon size="sm" color="white" />
-                    </Button>
+                    <CloseButton onClickHandler={onCropCancel} color="white"/>
```
DanielSchiavini: nice

src/components/image_upload.tsx
```diff
@@ -50,6 +50,8 @@ export interface ImageUploadState {
     dropActive: boolean;
     /** The URL the user started cropping with. This is used to keep the whole image even then the imageURL changed **/
     cropURL: string | null;
+    /** Whether the component is focused using the keyboard */
+    dropFocus: boolean;
```
DanielSchiavini: nice!

src/stories/form.stories.tsx
```diff
@@ -244,7 +247,7 @@ stories.add('Image upload', withState({ value: null })
                     id="id_8"
                     value={store.state.value}
                     onChangeHandler={(id, value: ImageFile) => store.set({ value: value.preview })}
-                    onReadHandler={action('read')}
+                    onReadHandler={(id, value: string) => store.set({ value })}
```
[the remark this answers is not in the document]
DanielSchiavini: your IDE has an invalid cache, the callback changed

src/components/image_cropper.tsx
DanielSchiavini: we don't

src/components/image_upload.tsx
```diff
@@ -30,81 +38,125 @@ export interface ImageUploadProps {
...
             const reader = new FileReader();
-            reader.onload = () => onFileRead(file, reader.result as string);
-            reader.onerror = console.error; // TODO
+            reader.onload = () => onFileRead(reader.result as string);
+            reader.onerror = console.error;
             reader.readAsDataURL(file);
         }
+
+        if (allowCropping) {
+            return this.setState({ cropURL: file.preview });
```
DanielSchiavini:
```suggestion
            this.setState({ cropURL: file.preview });
```

src/components/image_cropper.tsx
DanielSchiavini:
```suggestion
        imageURL: null,
```

PR #73, opened 2018-10-10
Adding min/max width/height to textareas

DanielSchiavini:
as requested in https://docs.google.com/presentation/d/1RAURl-eRmsAZUxA1bNGf2kNidV5BvfNdVs5s5m2OeO4/edit?ts=5bb7216f#slide=id.g43d723e7e9_1_80

### DIFF
--- a/src/components/input_fields/textarea_field.tsx
+++ b/src/components/input_fields/textarea_field.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 
 import InputFieldWrapper, {InputFieldProps} from './input_field_wrapper';
 import Textarea from '../textarea';
+import splitTextByNewLine from '../../utils/split_text_by_new_line'
 
 export interface TextareaFieldProps extends InputFieldProps {
     /** The text **/
@@ -47,7 +48,7 @@ class TextareaField extends React.Component<TextareaFieldProps, TextareaFieldSta
         const {id, staticField} = wrapperProps;
         const {touched} = this.state;
         return <InputFieldWrapper touched={touched} inputChildren={children} {...wrapperProps}>{
-            staticField ? value :
+            staticField ? splitTextByNewLine(value) :
             <Textarea
                 id={id}
                 onChangeHandler={this._onChange}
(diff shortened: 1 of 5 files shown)


src/components/input_fields/textarea_field.tsx
```diff
@@ -3,6 +3,7 @@ import * as React from 'react';
 
 import InputFieldWrapper, {InputFieldProps} from './input_field_wrapper';
 import Textarea from '../textarea';
+import splitTextByNewLine from '../../utils/split_text_by_new_line'
```
woosie10: semi-colon

src/components/input_fields/textarea_field.tsx
```diff
@@ -47,7 +48,7 @@ class TextareaField extends React.Component<TextareaFieldProps, TextareaFieldSta
         const {id, staticField} = wrapperProps;
         const {touched} = this.state;
         return <InputFieldWrapper touched={touched} inputChildren={children} {...wrapperProps}>{
-            staticField ? value :
+            staticField ? splitTextByNewLine(value) :
```
woosie10: splitTextByNewLine does not handle empty lines

src/stylesheets/_forms.scss
```diff
@@ -236,7 +236,10 @@
         @extend %input-style;
         vertical-align: top;
         height: auto;
+        min-width: 15rem;
+        max-width: 100%;
         min-height: 9.2rem;
+        max-height: 30rem;
```
woosie10: Lets change this to 100vh instead of 30rem. If it needs to be less in a specific implementation, then it can be done in the css of that application.

src/utils/split_text_by_new_line.tsx
```diff
@@ -6,7 +6,7 @@ import * as React from 'react';
  */
 function splitTextByNewLine(text: string): JSX.Element[] {
     if (text) {
-        return text.split("\n").map((p, i) => <p key={i}>{p}</p>);
+        return text.split("\n").map((p, i) => <p key={i}>{p || "&nbsp;"}</p>);
```
woosie10: This just prints out `&nbsp;` rather than adding a space. Also, we should remove the margin-bottom from the p for the textarea component

PR #77, opened 2018-10-16
Add multi-step functionality to the popup component

woosie10:
The PR adds multi-step functionality and a size prop to the popup component

### DIFF
--- a/src/components/__tests__/popup.test.tsx
+++ b/src/components/__tests__/popup.test.tsx
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
 
 // component
 import Popup from '../popup';
+import ProgressBar from '../progress_bar';
 
 describe('The Popup component', () => {
     let props;
@@ -44,14 +45,19 @@ describe('The Popup component', () => {
     });
 
     it('should reset primary button spinner when there are validation errors', () => {
-        wrapper.setState({primaryBtnShowSpinner: true});
-        wrapper.setProps({validationErrors: {}})
+        wrapper.setState({ primaryBtnShowSpinner: true });
+        wrapper.setProps({ validationErrors: {} })
         expect(wrapper.state('primaryBtnShowSpinner')).toBe(false);
     });
 
     it('should reset secondary button spinner when there are validation errors', () => {
-        wrapper.setState({secondaryBtnShowSpinner: true});
-        wrapper.setProps({validationErrors: {}})
+        wrapper.setState({ secondaryBtnShowSpinner: true });
+        wrapper.setProps({ validationErrors: {} })
         expect(wrapper.state('secondaryBtnShowSpinner')).toBe(false);
+    });
+
+    it('should render a progress bar for multi-step popups', () => {
+        wrapper.setProps({ step: 1, totalSteps: 3 });
+        expect(wrapper.find(ProgressBar)).toHaveLength(1);
     });
 });
--- a/src/components/__tests__/popup.test.tsx
+++ b/src/components/__tests__/popup.test.tsx
@@ -46,13 +46,13 @@ describe('The Popup component', () => {
 
     it('should reset primary button spinner when there are validation errors', () => {
         wrapper.setState({ primaryBtnShowSpinner: true });
-        wrapper.setProps({ validationErrors: {} })
+        wrapper.setProps({ validationErrors: {} });
         expect(wrapper.state('primaryBtnShowSpinner')).toBe(false);
     });
 
     it('should reset secondary button spinner when there are validation errors', () => {
         wrapper.setState({ secondaryBtnShowSpinner: true });
-        wrapper.setProps({ validationErrors: {} })
+        wrapper.setProps({ validationErrors: {} });
         expect(wrapper.state('secondaryBtnShowSpinner')).toBe(false);
     });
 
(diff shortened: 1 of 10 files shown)


src/components/popup.tsx
```diff
@@ -29,18 +30,24 @@ export interface PopupProps {
...
+    /** The current step number of a multi-step popup */
+    step?: number;
+    /** The total number of steps of a multi-step popup */
+    totalSteps?: number;
 }
 
 export interface PopupState {
     primaryBtnShowSpinner: boolean;
     secondaryBtnShowSpinner: boolean;
+    storedStep: number;
```
DanielSchiavini: I guess we need some explanation here on why we need to keep a copy of the prop

src/components/popup.tsx
```diff
@@ -74,37 +89,44 @@ export class Popup extends React.Component<PopupProps, PopupState> {
...
-                        secondaryBtnHandler={this._secondaryBtnHandler}
-                        secondaryBtnShowSpinner={secondaryBtnShowSpinner}
-                        validationErrors={validationErrors}
-                        alwaysEnableSubmitButton={true}
-                    >
-
-                        {children}
-
-                    </Form>
+        return <div className="popup">
```
DanielSchiavini: Will this div be positioned in a random place in the page, since I assume the `Modal` is positioned absolute or static?

src/stories/modal.stories.tsx
```diff
@@ -70,6 +71,63 @@ function validation(quantity: string | number): string {
...
+        case 1:
+            return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+        case 2:
+            return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.'
+        case 3:
+            return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec efficitur lectus lorem, id consectetur tellus sollicitudin eu. Curabitur nec metus nibh. Sed sagittis mi nisi, sollicitudin viverra magna malesuada nec. Morbi porttitor, nunc nec mollis feugiat, ex felis commodo sapien, vitae porta sapien magna sit amet nibh. Fusce quis porttitor massa. Pellentesque ornare risus at elementum tempor. Sed convallis odio tellus. Aenean tincidunt purus ut nisl ultrices, sit amet auctor odio porta. Mauris euismod ligula luctus semper dictum. Nulla orci leo, facilisis ac dapibus vitae, consequat ut sem.'
+    }
+}
+
+function getPrimaryBtnText(step: number): string {
```
DanielSchiavini: These 3 functions could be a default implementation, since I think they would otherwise be repeated. Of course they would have to accept any number of steps and be translated..

src/components/popup.tsx
woosie10: The popup class has no properties, it is only added here to scope the other classes. I like to do this for all components

src/stories/modal.stories.tsx
woosie10: I think the current solution gives the most flexibility

src/components/__tests__/popup.test.tsx
```diff
@@ -44,14 +45,19 @@ describe('The Popup component', () => {
     });
 
     it('should reset primary button spinner when there are validation errors', () => {
-        wrapper.setState({primaryBtnShowSpinner: true});
-        wrapper.setProps({validationErrors: {}})
+        wrapper.setState({ primaryBtnShowSpinner: true });
+        wrapper.setProps({ validationErrors: {} })
```
DanielSchiavini: missing `;`

src/components/__tests__/popup.test.tsx
```diff
@@ -44,14 +45,19 @@ describe('The Popup component', () => {
...
         expect(wrapper.state('primaryBtnShowSpinner')).toBe(false);
     });
 
     it('should reset secondary button spinner when there are validation errors', () => {
-        wrapper.setState({secondaryBtnShowSpinner: true});
-        wrapper.setProps({validationErrors: {}})
+        wrapper.setState({ secondaryBtnShowSpinner: true });
+        wrapper.setProps({ validationErrors: {} })
```
DanielSchiavini: missing `;`

src/components/popup.tsx
```diff
@@ -39,11 +41,16 @@ export interface PopupProps {
...
     secondaryBtnShowSpinner: boolean;
+    /** Saved solely to be used as a comparison in getDerivedStateFromProps to check whether the step prop has changed */
     storedStep: number;
 }
 
 export class Popup extends React.Component<PopupProps, PopupState> {
 
+    static defaultProps = {
+        width: 'sm'
+    }
```
DanielSchiavini: missing `;`

woosie10: it's a regular object, so you can't add a semi-colon

DanielSchiavini: you most definitely can: https://github.com/Ultimaker/react-web-components/blob/60094b1a5d8b2942160a4ac7c3cf41989df0eaa0/src/components/input_fields/input_field_wrapper.tsx#L72

woosie10: oh, I misunderstood what you meant

PR #80, opened 2018-10-23
Changes for cloud 2fa story

woosie10:


src/stylesheets/_input_fields.scss
```diff
@@ -78,6 +78,13 @@
         }
     }
 
+    &.input-field--textarea {
+        .static-field p {
```
DanielSchiavini: This should be applied to anything returned from splitTextByNewLine

PR #81, opened 2018-10-23
Making the popup and form components more flexible

DanielSchiavini:


ChrisTerBeke:
I think this is a good start towards extracting some things more in separate components.

PR #94, opened 2018-11-14
STAR-299: Static field

DanielSchiavini:
This PR adds a new input type that simply outputs its value and eventual children.
We can use this simplified version for fields that are always read-only anyway.

### DIFF
--- a/src/components/input_fields/__tests__/static_field.test.tsx
+++ b/src/components/input_fields/__tests__/static_field.test.tsx
@@ -25,7 +25,8 @@ describe('The static field component', () => {
     });
 
     it("should render it's value and its children", () => {
-        expect(wrapper.children().map(c => c.text())).toEqual(["2018", "child"]);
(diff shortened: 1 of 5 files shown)


DanielSchiavini:
Should I still add it to storybook?

src/components/input_fields/__tests__/static_field.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+        wrapper = shallow(<StaticField {...props}>child</StaticField>);
+    });
+
+    it('should render an input wrapper', () => {
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it("should render it's value and its children", () => {
+        expect(wrapper.children().text()).toEqual("2018");
+        expect(wrapper.find(InputFieldWrapper).prop('inputChildren')).toEqual("child");
```
woosie10: Quotes

woosie10:
And yes, please add to storybook

PR #101, opened 2018-11-23
Cl 1013 extend popup capabilities, add aspect ratio keeper component

ImreBognarUltimaker:


src/components/keep_aspect_ratio.tsx
```diff
@@ -0,0 +1,43 @@
...
+    aspectRatio?: number;
+    /** An optional id for the container **/
+    id?: string;
+    /** An optional class name for the container **/
+    className?: string;
+}
+
+/**
+ * returns only the first child with a modified className
+ * @param param0
```
woosie10: whats param0 ?

src/components/keep_aspect_ratio.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import classNames = require('classnames');
+
+export interface keepAspectRatioProps {
+    /** the content to be sized appropriately  */
+    content: JSX.Element;
```
woosie10: would be better to pass in the element as the children

src/components/keep_aspect_ratio.tsx
```diff
@@ -0,0 +1,43 @@
...
+    id?: string;
+    /** An optional class name for the container **/
+    className?: string;
+}
+
+/**
+ * returns only the first child with a modified className
+ * @param param0 
+ */
+export const keepAspectRatio: React.StatelessComponent<keepAspectRatioProps> = ({ aspectRatio, className, id, content }) => {
```
woosie10: Would a name like aspectRatioContainer be better?

src/stories/other.stories.tsx
```diff
@@ -77,3 +78,14 @@ stories.add('Circle Icon', withInfo(
         {text('Text', '1')}
     </CircleIcon>
 ));
+
+stories.add('Keep aspect ratio', withInfo(
+    'A container to keep a given aspect ratio for an image or iframe'
+)(() =>
+    <div>
+        <div style={{ width: text('Width', '40rem'), backgroundColor: '#DDD'}} >
```
woosie10: Don't add knobs that are not props to the component as it's confusing

src/stories/other.stories.tsx
```diff
@@ -77,3 +78,14 @@ stories.add('Circle Icon', withInfo(
         {text('Text', '1')}
     </CircleIcon>
 ));
+
+stories.add('Keep aspect ratio', withInfo(
```
woosie10: can you please move this story to the layout stories

src/stories/other.stories.tsx
```diff
@@ -77,3 +78,14 @@ stories.add('Circle Icon', withInfo(
...
+    'A container to keep a given aspect ratio for an image or iframe'
+)(() =>
+    <div>
+        <div style={{ width: text('Width', '40rem'), backgroundColor: '#DDD'}} >
+            <KeepAspectRatio aspectRatio={(text('Aspect ratio', 16 / 9))} className='my-picture-frame' content={<iframe className='interesting-video' width="560" height="315" src="https://www.youtube.com/embed/160yAufQ-is" allowFullScreen></iframe>} >
```
woosie10: Can you just use a grey rectangle instead of the youtube video to simplify what this is showing

src/components/keep_aspect_ratio.tsx
ImreBognarUltimaker: I agree. I tried it, didn't work well unfortunately.
I would have to add a class to the child, to avoid using a descendant selector in CSS. It worked, but was very easy to break by accident, and could produce unexpected results.

src/components/keep_aspect_ratio.tsx
ImreBognarUltimaker: this comment belongs to a previous iteration, updated it

src/stories/other.stories.tsx
ImreBognarUltimaker: okay, removed it

src/components/keep_aspect_ratio.tsx
ImreBognarUltimaker: Your english is better, so I renamed it.

src/components/keep_aspect_ratio.tsx
ImreBognarUltimaker: Changed it to work like that.

src/stories/layout.stories.tsx
```diff
@@ -194,3 +195,15 @@ stories.add('Setting Panel', withState({ value: null })
         )
     )
 );
+
+stories.add('Keep aspect ratio', withInfo(
```
woosie10: This name needs updating

src/stories/layout.stories.tsx
```diff
@@ -194,3 +195,15 @@ stories.add('Setting Panel', withState({ value: null })
...
+    'A container to keep a given aspect ratio for an image or iframe'
+)(() =>
+    <div>
+        <div style={{ width: '40rem', backgroundColor: '#DDD' }} >
```
woosie10: Can we remove this div and just use the div inside the AspectRatioContainer. Also, div inside the AspectRatioContainer be so that it resizes when you make the screen smaller, i.e. so that it is responsive

PR #102, opened 2018-11-27
STAR-326: Allowing more responsivity in carousel

DanielSchiavini:


woosie10:
One other thing, can we please make the dots below the carousel blue like the market place, as that is our default colour

src/components/carousel.tsx
```diff
@@ -28,7 +33,19 @@ export const Carousel: React.StatelessComponent<CarouselProps> = ({ children, it
     // create an object with the format {breakpointWidth: {items: item_count}}.
     const responsive = _.zipObject(BreakpointSizes.slice(0, itemCounts.length), itemCounts.map(items => ({items})));
 
-    // each child will receive the given extra parameter
+    const breakpoint = Math.max(...BreakpointSizes.filter(bp => bp <= innerWidth && responsive[bp]));
+    if (responsive[breakpoint].items > React.Children.count(children)) {
```
woosie10: I think this should be `>=` rather than >

src/components/carousel.tsx
```diff
@@ -28,10 +33,15 @@ export const Carousel: React.StatelessComponent<CarouselProps> = ({ children, it
...
+    // renders a grid without any carousel features, in case we don't have enough items in the carousel.
+    const renderGrid = () =>
+        <Grid align="center" className="carousel__fixed">
+            {React.Children.map(children, child => React.isValidElement(child) &&
+                <GridItem layoutWidth="fit">{child}</GridItem>)
+            }
+        </Grid>;
 
-    return (
+    const renderCarousel = () =>
```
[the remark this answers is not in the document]
DanielSchiavini: I like to avoid the brackets so the alignment is good. How about parenthesis? Then we don't need the return.

src/components/carousel.tsx
```diff
@@ -40,9 +50,14 @@ export const Carousel: React.StatelessComponent<CarouselProps> = ({ children, it
...
-            {React.Children.map(children, child => React.isValidElement(child) && React.cloneElement(child, extraProps))}
-        </AliceCarousel>
-    );
+            {React.Children.map(children, child =>
+                React.isValidElement(child) && React.cloneElement(child, {onDragStart: e => e.preventDefault()} as any))
+            }
+        </AliceCarousel>;
+
+    // choose which
+    const breakpoint = Math.max(...BreakpointSizes.filter(bp => bp <= innerWidth && responsive[bp]));
```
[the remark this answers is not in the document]
DanielSchiavini: sure

src/stylesheets/_layout.scss
```diff
@@ -316,41 +316,25 @@
...
 
-.show-sm {
-    display: block;
-
-    @media (min-width: #{map-get($layout-breakpoints, md)}) {
-        display: none !important;
-        visibility: hidden !important;
-    }
-}
+@include layout-hiding-rules($layout-breakpoints);
```
woosie10: .hide-xs and .show-xs will now be missing as xs is not in $layout-breakpoints

woosie10: quick fix is to also add `@include layout-hiding-rules((xs: 0));`

woosie10: I've just added in the suggestion above, as I needed to rebase on this branch and this broke what I was doing

DanielSchiavini: Cool, so is the PR ok now?

src/components/carousel.tsx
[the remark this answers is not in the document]
DanielSchiavini: This is already done..?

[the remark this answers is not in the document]
DanielSchiavini: This is not applicable anymore

PR #113, opened 2018-12-12
STAR-348: Adding error popup for images that are too large

DanielSchiavini:


src/components/button.tsx
```diff
@@ -124,7 +128,7 @@ export class Button extends React.Component<ButtonProps, {}> {
         if (type === 'link' && this._isLinkInternal()) {
             return this._renderInternalLink(classes);
         }
-        if (type === 'link') {
+        if (type === 'link' && !disabled && !showSpinner && linkURL) {
```
woosie10: This is going to hide the link if it is disabled. I don't think you want that

DanielSchiavini: It shows a button then, because if linkURL is not a string the `href` prop is incorrect

PR #116, opened 2018-12-19
Refactor all drop down menus to use DropDownMenuBase

woosie10:
This PR refactors all drop down menus to use the same base component.


src/stories/menu.stories.tsx
```diff
@@ -7,6 +7,7 @@ import {
 import styles from '@sambego/storybook-styles';
 import { withInfo } from '@storybook/addon-info';
 
+import { relative } from 'path';
```
ImreBognarUltimaker: This import is not used.

PR #120, opened 2018-12-20
Rename style to appearance

woosie10:
To avoid names clashing, the attributte 'style' should only be used for in-line styling. All other uses will be changed to 'appearance'.

### DIFF
--- a/src/components/button.tsx
+++ b/src/components/button.tsx
@@ -6,7 +6,7 @@ import { Link } from 'react-router-dom';
 import Spinner from './spinner';
 
 export type ButtonType = 'submit' | 'button' | 'link';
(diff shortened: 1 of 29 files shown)


src/components/button.tsx
```diff
@@ -6,7 +6,7 @@ import { Link } from 'react-router-dom';
 import Spinner from './spinner';
 
 export type ButtonType = 'submit' | 'button' | 'link';
-export type ButtonStyle = 'primary' | 'secondary' | 'quiet' | 'alert' | 'link' | 'no-style';
+export type ButtonAppearance = 'primary' | 'secondary' | 'quiet' | 'alert' | 'link' | 'no-style';
```
DanielSchiavini: no-style is a bit weird now, shouldn't it be `'none'` or `null`?

PR #123, opened 2018-12-21
only add onCloseMenuHandler to child components that need it

woosie10:


src/components/drop_down_menu_base.tsx
```diff
@@ -31,6 +31,19 @@ export class DropDownMenuBase extends React.Component<DropDownMenuBaseProps, {}>
         e.stopPropagation();
     }
 
+    static _isHandlerRequired(child: any): boolean {
```
DanielSchiavini: can we add some jsdoc to this?

PR #124, opened 2019-01-09
remove onCloseMenuHandler from the drop down menu

woosie10:
An onCloseMenuHandler was passed from DropDownMenuBase to DropDownMenuItem, but this is causing some issues in cura connect. To simplify things, closing the menu on item click will be removed. Closing the menu on click outside will still work.
CL-1207


ChrisTerBeke:
So this means that if I click a menu item the menu will stay open?

PR #127, opened 2019-01-18
close drop down menus on item click

woosie10:
Unlike the new version of the context menu, drop down menus (more specifically select menus) need to be uncontrolled. This PR adds closing the menu when a menu item is clicked.

### DIFF
--- a/src/components/drop_down_menu_item.tsx
+++ b/src/components/drop_down_menu_item.tsx
@@ -15,23 +15,32 @@ export interface DropDownMenuItemProps {
     id?: string;
     /** Optional class for the menu item */
     className?: string;
+    /** Passed by the DropDownMenu component */
+    onCloseMenuHandler?: () => void;
 }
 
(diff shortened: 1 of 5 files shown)


src/components/drop_down_menu_item.tsx
```diff
@@ -15,23 +15,32 @@ export interface DropDownMenuItemProps {
...
-            {children}
-        </Button>
-    </li>
-);
+export const DropDownMenuItem: React.StatelessComponent<DropDownMenuItemProps> = ({
+    active, disabled, onClickHandler, id, className, onCloseMenuHandler, children,
+}) => {
+    const _onClickHandler = () => {
+        onClickHandler();
+        onCloseMenuHandler();
```
ImreBognarUltimaker: When you call this, should there not be a check if it's passed to the component?
Like:
onCloseMenuHandler && onCloseMenuHandler()

PR #128, opened 2019-01-22
STAR-335 move tile styles

woosie10:


src/stylesheets/_forms_actions.scss
```diff
@@ -22,6 +22,6 @@
     justify-content: center;
 }
```
ImreBognarUltimaker: It would be nice to add a class here for .form-actions--right

woosie10: added

PR #135, opened 2019-02-11
STAR-427 smaller bundles

woosie10:
This PR does the following:
- swaps react-tag-input for react-tag-autocomplete
- swaps react-dates for react-date-picker
- removes lodash (but keeps lodash.debounce)
- removes NotoSans-Bold font


src/components/image_cropper.tsx
```diff
@@ -13,6 +12,8 @@ if ('default' in AvatarEditor) {
     AvatarEditor = AvatarEditor.default;
 }
 
+const debounce = require('lodash.debounce');
```
DanielSchiavini: Why do we need to import it like this?

src/components/input_fields/code_field.tsx
```diff
@@ -52,13 +50,20 @@ export default class CodeField extends React.Component<CodeFieldProps, CodeField
...
      * for re-initializing the array whenever {maxLength} changes.
      */
-    private readonly _inputRefs: HTMLInputElement[] = [];
 
     /** The default props */
     static defaultProps = {
         type: 'number',
     };
 
+    static range(start: number, end: number): number[] {
```
DanielSchiavini: we could have this in a separate file so it can be reused

src/components/image_cropper.tsx
woosie10: The normal import wasn't working, but I have figured at the correct one

src/components/input_fields/code_field.tsx
woosie10: moved

src/components/input_fields/__tests__/date_field.test.tsx
```diff
@@ -48,18 +52,24 @@ describe('The date field component', () => {
     it('should render a static date', () => {
         wrapper.setProps({ staticField: true });
         expect(wrapper.find(DatePicker)).toHaveLength(0);
-        expect(wrapper.find(InputFieldWrapper).prop('children')).toEqual('31-12-2018');
+        expect(wrapper.find(InputFieldWrapper).prop('children')).toEqual('1/1/1970');
```
DanielSchiavini: Why is this in US-format? That's may break the back end.

PR #137, opened 2019-02-13
Import the old way for now

woosie10:
To keep both parcel and webpack happy, lodash functions need to be imported this way for now.


ChrisTerBeke:
Maybe add a comment line about why this is needed?

PR #149, opened 2019-02-19
STAR-395 update the animation library

woosie10:
This PR removes the dependancies React-motion and React-collapse and replaces them with React-spring.


ImreBognarUltimaker:
The date_field.test.tsx is failing for me.

PR #159, opened 2019-03-18
Fix translation utility class

ChrisTerBeke:
* Change how translation file are loaded (to increase performance, only the active locale file is loaded).
* Fixes in the I18n class.

We should add more tests for i18n, but let's do that in a separate PR.

### DIFF
--- a/src/components/__tests__/response_error.test.tsx
+++ b/src/components/__tests__/response_error.test.tsx
@@ -6,14 +6,14 @@ import { shallow } from 'enzyme';
 import { I18n } from '../../utils/i18n';
 
 // component
-import ResponseError, { ErrorMessageTemplates, FieldTranslations } from '../response_error';
+import ResponseError from '../response_error';
 
-const ERROR_CODE_TO_I18NC_TEMPLATE: ErrorMessageTemplates = {
-    fieldError: I18n.translate('error message', '%{field_name} was invalid: %{title}.'),
+const ERROR_CODE_TO_I18NC_TEMPLATE = {
+    fieldError: () => I18n.translate('error message', '%{field_name} was invalid: %{title}.'),
 };
 
-const FIELD_NAME_TO_I18NC: FieldTranslations = {
-    username: I18n.translate('form field username', 'Username'),
+const FIELD_NAME_TO_I18NC = {
+    username: () => I18n.translate('form field username', 'Username'),
 };
 
 describe('The ResponseError component', () => {
--- a/src/components/response_error.tsx
+++ b/src/components/response_error.tsx
@@ -26,9 +26,9 @@ export const getTranslatedError = (
         Object.keys(error.meta).forEach((key) => {
             if (key === 'field_name' && fieldTranslations[params[key]]) {
                 // each field translation is a function that returns a string
-                params[key] = fieldTranslations[params[key]]()
+                params[key] = fieldTranslations[params[key]]();
             } else {
-                params[key] = error.meta[key]
+                params[key] = error.meta[key];
             }
         });
     }
@@ -38,7 +38,6 @@ export const getTranslatedError = (
     const errorMessage = errorMessageTemplates[error.code]
         ? errorMessageTemplates[error.code]()
         : I18n.translate('error message default', 'There was a problem on the server: %{title} (code %{code})');
-        
     return I18n.interpolate(errorMessage, params);
 };
 
--- a/src/components/response_error.tsx
+++ b/src/components/response_error.tsx
@@ -8,13 +8,6 @@ import Tile from './tile';
 // utils
 import { I18n } from '../utils/i18n';
 
-// a mapping of error codes to translated error messages which may include
-// the error parameters as replacement keys
-export type ErrorMessageTemplates = { [key: string]: string };
-
-// a mapping of field names to translated versions
-export type FieldTranslations = { [key: string]: string };
-
 /**
  * Gets a translated message for the given error.
  * @param error - The error object to be translated.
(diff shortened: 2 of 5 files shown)


src/components/response_error.tsx
```diff
@@ -8,13 +8,6 @@ import Tile from './tile';
 // utils
 import { I18n } from '../utils/i18n';
 
-// a mapping of error codes to translated error messages which may include
-// the error parameters as replacement keys
-export type ErrorMessageTemplates = { [key: string]: string };
```
DanielSchiavini: Why not change this to `{ [key: string]: () => string },` here? We can share this for both error messages and fields, e.g. 
`export type TranslationTemplates = { [key: string]: () => string }`

src/components/response_error.tsx
```diff
@@ -23,17 +16,28 @@ export type FieldTranslations = { [key: string]: string };
...
+    fieldTranslations: { [key: string]: () => string },
 ): string => {
     const params = { ...error };
+
+    // check if we need to translate any fields
     if (params.meta) {
         Object.keys(error.meta).forEach((key) => {
-            params[key] = (key === 'field_name' && fieldTranslations[params[key]]) || error.meta[key];
+            if (key === 'field_name' && fieldTranslations[params[key]]) {
+                // each field translation returns a function that returns a string
```
DanielSchiavini: `s/returns a function/is a function`

src/components/response_error.tsx
ChrisTerBeke: It's not used that often and this reduces a 'dependency' on an import from react-web-components from the generated files, which source code lives in commons (and thus can become outdated quickly).

src/components/__tests__/response_error.test.tsx
```diff
@@ -6,14 +6,14 @@ import { shallow } from 'enzyme';
...
 
-const ERROR_CODE_TO_I18NC_TEMPLATE: ErrorMessageTemplates = {
-    fieldError: I18n.translate('error message', '%{field_name} was invalid: %{title}.'),
+const ERROR_CODE_TO_I18NC_TEMPLATE = {
+    fieldError: () => I18n.translate('error message', '%{field_name} was invalid: %{title}.'),
 };
 
-const FIELD_NAME_TO_I18NC: FieldTranslations = {
-    username: I18n.translate('form field username', 'Username'),
+const FIELD_NAME_TO_I18NC = {
```
DanielSchiavini: Now we have no type checking here and in commons

src/components/response_error.tsx
DanielSchiavini: I do prefer to have it though otherwise we don't know when stuff is broken

src/utils/i18n.tsx
```diff
@@ -1,39 +1,34 @@
...
-        // load the translation files
-        translations.forEach(async translation => (
-            this._loadTranslation(translation.name, translation.source)
-        ));
+     * Initialize the Gettext context.
+     * @param translations The translation keys with files to load.
+     */
+    public static async initialize(translations: { [s: string]: string; }): Promise<void> {
+        // register supported languages
+        Object.keys(translations).forEach((language) => {
```
DanielSchiavini: Can't we use `Object.keys(translations).forEach(I18n._supportedLanguages.push)`?

src/utils/i18n.tsx
```diff
@@ -1,39 +1,34 @@
...
+            I18n._supportedLanguages.push(language);
+        });
 
         // set the default locale based on the browser settings
-        I18n._gt.setLocale(I18n.getLocale());
+        const locale = I18n.getLocale();
+        if (I18n.isLanguageSupported(locale)) {
```
DanielSchiavini: Shouldn't we have a default language when not supported? Also can't we have multiple locale in the browser? I have for example en-US, nl-NL, pt-BR, en-UK, nl-BE, pt-PT configured in this order

src/utils/i18n.tsx
```diff
@@ -58,8 +53,8 @@ export class I18n {
     * Check if the given language is supported.
     * @param language The language to check, for example 'en'.
     */
-    public static isLanguageSupported(language: Languages): boolean {
-        return I18n._supportedLanguages.indexOf(language) > -1;
+    public static isLanguageSupported(language: string): boolean {
+        return I18n._supportedLanguages.indexOf(language) > -1 && language !== undefined;
```
DanielSchiavini: I18n._supportedLanguages.indexOf(undefined) == -1 right, why do we need `language !== undefined`?
Note comparing to undefined doesn't work always, you need to use [typeof language == 'undefined'](https://stackoverflow.com/questions/2778901/how-to-compare-variables-to-undefined-if-i-don-t-know-whether-they-exist)

src/components/__tests__/response_error.test.tsx
ChrisTerBeke: There's type checking on the function itself though, this is only the test input. If it's wrong, your editor will indicate that the type doesn't match the expected type.

src/utils/i18n.tsx
ChrisTerBeke: That would be cleaner indeed.

ChrisTerBeke: nvm, the function signatures of forEach and push don't match, so typescript doesn't allow it.

src/utils/i18n.tsx
ChrisTerBeke: The default language is english, this is already handled by the gettext library itself when registering translations. It support multi-locale fine, if there's multiple it just gets the top one from navigator.languages.

src/utils/i18n.tsx
ChrisTerBeke: You're right, that slipped through when i was testing.

ChrisTerBeke:
@ImreBognarUltimaker can you confirm?

src/utils/i18n.tsx
sedwards2009: Comparing to `undefined` is fine here. TypeScript should prevent you from trying to access a variable which hasn't been defined, which is what Stack Overflow was talking about.

DanielSchiavini: `undefined` is not a keyword in JS, if any library does `undefined=1` this suddenly breaks. It shouldn't happen though..

sedwards2009: The global `undefined` variable has been read only since the Firefox 4 and ES5 era. You can't overwrite it.

DanielSchiavini: Didn't know that, thanks!

PR #161, opened 2019-03-22
Remove unused fields and log

DanielSchiavini:


DanielSchiavini:
They are called by other apps, but we always subclass the NavRoute interface

PR #168, opened 2019-04-04
Resource grid components

woosie10:
This PR moves the resource grid components from MyCloud and improves them. It also adds a resource grid story to Storybook, as well as tests.


src/stylesheets/_resource_empty_tile.scss
```diff
@@ -0,0 +1,15 @@
+.resource-empty-state {
+    text-align: center;
+
+    .resource-empty-state__text {
```
ImreBognarUltimaker: Must these be nested?

src/stylesheets/_resource_empty_tile.scss
```diff
@@ -0,0 +1,15 @@
+.resource-empty-state {
```
ImreBognarUltimaker: Shouldn't this file be named _resource_empty_state.scss ?

src/stylesheets/_sub_resource_list.scss
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    .pill {
+        height: 3rem;
+        line-height: 3rem;
+    }
+
+    .more-pill .pill {
+        .more-pill__text {
+            display: inline-block;
```
ImreBognarUltimaker: I think using only .more-pill__text here should be enough.

src/stylesheets/_resource_empty_tile.scss
woosie10: updated

src/stylesheets/_resource_empty_tile.scss
woosie10: updated

src/stylesheets/_sub_resource_list.scss
woosie10: updated

PR #182, opened 2019-04-23
STAR-519 new header

woosie10:
This PR adds a new version of the header and navigation components, and renames the old ones.

### DIFF
--- a/src/components/__tests__/header.test.tsx
+++ b/src/components/__tests__/header.test.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 // component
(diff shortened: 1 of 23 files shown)


src/components/__tests__/header.test.tsx
```diff
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
```
ImreBognarUltimaker: It would be nice to up the function coverage on header into green.

woosie10: Coverage for the header is now 100%

ImreBognarUltimaker: Prima.

PR #187, opened 2019-05-01
Fix disabling scroll for slide in panel and modal

woosie10:


src/components/__tests__/modal.test.tsx
```diff
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { shallow, mount } from 'enzyme';
 
 // component
+import { loadavg } from 'os';
```
ImreBognarUltimaker: I think this is not used.

woosie10: I've removed this now

PR #193, opened 2019-05-13
Fix showing image cropper

RoelWKramer:
this is probably the way it should be fixed, see other PR for a fix which does not break the testing logic.

See other PR: https://github.com/Ultimaker/react-web-components/pull/194

Which solution is the best?

### DIFF
--- a/src/components/image_upload.tsx
+++ b/src/components/image_upload.tsx
@@ -113,11 +113,6 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
         };
         reader.onerror = console.error;
         reader.readAsDataURL(file);
-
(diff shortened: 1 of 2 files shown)


src/components/image_upload.tsx
```diff
@@ -102,16 +100,24 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
...
+                onFileRead(reader.result as string);
+            }
+        };
+        reader.onerror = console.error;
+        reader.readAsDataURL(file);
+
+        // // This seems only to be used in test cases, as file.preview is not available
+        // if (allowCropping && file.preview !== null) {
+        //     this.setState({ cropURL: file.preview });
+        // }
```
woosie10: Lets just remove this

PR #194, opened 2019-05-13
Fix showing image cropper, in a lesser breaking way

RoelWKramer:
keeps more logic like it was, see PR https://github.com/Ultimaker/react-web-components/pull/193 for the other solution.

### DIFF
--- a/src/components/image_upload.tsx
+++ b/src/components/image_upload.tsx
@@ -88,11 +88,9 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
     private _onDropHandler(files: ImageFile[]): void {
         const file = files[0];
         this.setState({ dropActive: false });
-
         const {
             allowCropping, onFileSelection, onFileRead, maxMb,
         } = this.props;
-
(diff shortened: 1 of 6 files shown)


src/components/image_upload.tsx
```diff
@@ -104,12 +102,20 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
...
+                if (onFileRead) {
+                    onFileRead(reader.result as string);
+                }
+            };
             reader.onerror = console.error;
             reader.readAsDataURL(file);
         }
 
-        if (allowCropping) {
+        // This seems only to be used in test cases, as file.preview is not available in the browser
```
ChrisTerBeke: That seems weird, I think we use preview for showing the preview while cropping?

RoelWKramer: Nope! Preview is there, with this solution as well, as far as I can see.
Also makes no sense to crop the preview image. You want to crop the original image?

RoelWKramer: This is why i think the other solution is better, but this keeps everything compatible.

PR #198, opened 2019-05-21
Add npm publish on new tags

rvdh:


ChrisTerBeke:
@woosie10 can you test if this actually works?

PR #200, opened 2019-05-23
Set variables so git can commit

rvdh:


.gitlab-ci.yml
```diff
@@ -1,5 +1,11 @@
 image: node:8
 
+variables:
```
ChrisTerBeke: Maybe better to do this in a before_script so you don't have to wait for the end of the CI process to see if it works?

PR #204, opened 2019-05-27
STAR-577 add disabled option to the context menu

ImreBognarUltimaker:
STAR-577
Add a disabled option to the context menu component.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -29,7 +29,27 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.1.6", "@babel/core@^7.2.2":
+"@babel/core@^7.1.0":
+  version "7.4.5"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -29,7 +29,27 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.1.6", "@babel/core@^7.2.2":
+"@babel/core@^7.1.0":
```
ChrisTerBeke: Why did so many dependencies change? A diff of 1500+ lines.

ImreBognarUltimaker: This branch is based on this PR:
https://github.com/Ultimaker/react-web-components/pull/203

Large libraries were updated.

ChrisTerBeke: It's hard to see when basing branches on other branches than master. It's better to leave those completely separate PRs (as you cannot merge this branch now until the other one is unless you accept all those updates as well).

ImreBognarUltimaker: I will make PRs like this a draft in the future, until the one on which it's based on is merged.

PR #207, opened 2019-05-28
STAR-620 add sub navigation

woosie10:
This PR adds sub navigation as well as refactoring the existing navigation into smaller components.

The design is based on the following design, but kept more inline with our current style.
https://app.zeplin.io/project/5b18f693e66f3e4636a2a54a/screen/5cdea856407b951d80b99613


src/components/__tests__/__snapshots__/navigation_desktop.test.tsx.snap
```diff
@@ -0,0 +1,121 @@
...
+exports[`The NavigationDesktop component should render 1`] = `
+<div
+  className="navigation__desktop show-sm layout layout--gutter-none layout--align-middle"
+>
+  <div
+    className="layout__item u-fit-sm"
+  >
+    <nav
+      className="navigation__container"
+      role="navigation"
```
ImreBognarUltimaker: It is not necessary to add navigation role to a nav element.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

woosie10: updated

src/components/navigation_mobile.tsx
```diff
@@ -0,0 +1,187 @@
...
+    constructor(props) {
+        super(props);
+
+        this._onCloseMobileMenuHandler = this._onCloseMobileMenuHandler.bind(this);
+    }
+
+    private _onCloseMobileMenuHandler(): void {
+        const { onCloseMobileMenuHandler } = this.props;
+
+        if (onCloseMobileMenuHandler) {
```
ImreBognarUltimaker: The else of this if is not covered by the tests.

woosie10: added extra test

PR #224, opened 2019-07-12
CS-125 Fix shifting header items

ianpaschal:
I've added some utils to actually block mouse scroll input programatically.

Why didn't I do it in CSS?

Because I see doing it in CSS as a hack. What we truly want is to control when to ignore scroll input, namely based on component props. Using a `noscroll` sort of makes it look like that's what's happening, but of course leads to other bugs, such as the fact that the page height is overridden and then scroll bars disappear. Hence: hacky.

So after some research I found this to be the most stable solution which works for desktop and mobile and truly does what we actually want (as I see it).

### DIFF
--- a/src/utils/toggle_scrolling.tsx
+++ b/src/utils/toggle_scrolling.tsx
@@ -1,8 +1,16 @@
 // Adapted from from https://stackoverflow.com/a/4770179
 
-// spacebar: 32, pageup: 33, pagedown: 34, end: 35, home: 36
-// left: 37, up: 38, right: 39, down: 40,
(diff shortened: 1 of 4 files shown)


ChrisTerBeke:
I agree that events like keyboard or mouse input should be controlled by JavasScript, not CSS.

src/utils/toggle_scrolling.tsx
```diff
@@ -0,0 +1,68 @@
+// Adapted from from https://stackoverflow.com/a/4770179
```
ImreBognarUltimaker: Please make some test for this.

PR #228, opened 2019-07-23
Set auth token within .gitlab-ci.yml

ianpaschal:
This is a better solution because the inclusion of a `.npmrc` file meant adding or removing any packages with yarn threw an error and failed. It's true a dev could just delete that file but they'd have to make sure they never accidentally committed that change, and requiring the deletion of config to run `yarn install` is uncool.


ChrisTerBeke:
Good one!

PR #229, opened 2019-07-23
Correctly expose IconWrapper and IconProps

ianpaschal:
Contributes to CS-180


src/index.tsx
```diff
@@ -106,8 +106,7 @@ export { default as DownloadIcon } from './components/icons/download_icon';
 export { default as DraftIcon } from './components/icons/draft_icon';
 export { default as EmailIcon } from './components/icons/email_icon';
 export { default as HighlightIcon } from './components/icons/highlight_icon';
-export { default as IconWrapper } from './components/icons/icon_wrapper';
-export { default as IconProps } from './components/icons/icon_wrapper';
+export * from './components/icons/icon_wrapper';
```
ImreBognarUltimaker: It is alright, but could you please put it at the front of this icon section?

ianpaschal: Done. To be honest I don't think it should be in the icon section at all but I'm afraid of breaking things by moving the file to the main components section.

ImreBognarUltimaker: It should be safe to do that.

ianpaschal: Ok let me see if it i can do it without breaking every import statement

PR #231, opened 2019-08-23
CS-220 add new icon

ImreBognarUltimaker:
CS-220

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-    "version": "5.6.0",
+    "version": "5.5.2",
     "name": "@ultimaker/react-web-components",
     "description": "Ultimaker's unified react component and style library for front-end web",
     "main": "./dist/index.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "5.5.2",
+    "version": "5.6.0",
```
ChrisTerBeke: There's no need to change the version number in such a PR. One a tag is made the CI will commit the changed version number back to the repository automatically.

src/components/icons/info_circular_icon.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { IconWrapper, IconProps } from './icon_wrapper';
+
+/** this icon works a bit differently than the rest
+ * when it's color is modified, it's affecting the 'i' letter in the inside
```
ChrisTerBeke: sentences are off by 1 space.

PR #236, opened 2019-08-29
Replace Ultimaker logo with text

ianpaschal:
This includes lots of fonts. Most of them are commented out so the browser won't load them, but I included them to keep things standardized so if we ever _do_ need them, we have them in the same version/character set/etc.


ianpaschal:
Also, I keep them under a single font family and use weights instead, which is more... kosher... in my opinion.

ianpaschal:
Screen Shot 2019-08-29 at 11 43 23

PR #244, opened 2019-09-17
Replace node-sass with Dart sass

matijs:
- use sass instead of node-sass
- bring `.storybook/webpack.config.js` in line with code style


ImreBognarUltimaker:
Why is it better?

matijs:
I believe Dart Sass is now the reference implementation. It's also slightly stricter but most importantly, there are no requirements for platform specific binaries as it's all JS.

matijs:
(I am also trying to speed up things a little. Having to wait ~3 minutes for Storybook to start is not really acceptable)

ChrisTerBeke:
Always nice to see improvements :) I have no opinion on node-sass vs sass, so if the front-enders are happy with its let's do it.

ChrisTerBeke:
Approved by multiple people, I'll just merge it I guess to not have lingering PRs.

PR #245, opened 2019-09-17
Make people do npm version manually

rvdh:


ChrisTerBeke:
I guess @matijs wants to have a look as it was flagged by him?

ianpaschal:
Thanks @rvdh! I think this is a pretty good solution.

rvdh:
> Would this require to push directly to master (the version number in package.json)? If so, that's gonna be a problem.

Dont think so, npm will just commit to whatever branch you're on. But that does mean you'll have to create a PR to merge the version update in package.json and the tag will point to a commit in a branch.

ChrisTerBeke:
> > Would this require to push directly to master (the version number in package.json)? If so, that's gonna be a problem.
> 
> Dont think so, npm will just commit to whatever branch you're on. But that does mean you'll have to create a PR to merge the version update in package.json and the tag will point to a commit in a branch.

That's fine, that commit will end up in master after merging anyways.

ChrisTerBeke:
Btw, also needed in stardust-web-components and that new eslint repo?

PR #246, opened 2019-09-17
Update `<BoxPlaceholder />`: allow choosing background colours

xli12:
* According to [this requirement](https://jira.ultimaker.com/browse/CS-256?focusedCommentId=90624&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-90624) and [this requirement](https://jira.ultimaker.com/browse/CS-256?focusedCommentId=90624&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-90624) the box placeholder can be grey or white (transparent); adding the functionality to enable it. Default: transparent.

### DIFF
--- a/src/components/__tests__/box_placeholder.test.tsx
+++ b/src/components/__tests__/box_placeholder.test.tsx
@@ -20,9 +20,9 @@ describe('The BoxPlaceholder component', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
-    it('should render with className "background--grey"', () => {
-        props.backgroundColor = 'grey';
+    it('should render with className "background--light"', () => {
+        props.backgroundColor = 'light';
         wrapper.setProps(props);
-        expect(wrapper).toMatchSnapshot();
+        expect(wrapper.hasClass('background--light')).toBe(true);
     });
 });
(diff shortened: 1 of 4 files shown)


src/components/__tests__/box_placeholder.test.tsx
```diff
@@ -16,7 +16,13 @@ describe('The BoxPlaceholder component', () => {
         wrapper = shallow(<BoxPlaceholder {...props} />);
     });
 
-    it('should render', () => {
+    it('should render with className "background--transparent"', () => {
+        expect(wrapper).toMatchSnapshot();
```
ChrisTerBeke: Maybe a whole snapshot is a bit overkill here and we just wanna test the css class property for it's value?

src/components/box_placeholder.tsx
```diff
@@ -2,21 +2,31 @@ import * as React from 'react';
 
 import BoxIcon from './icons/box_icon';
 
+export type BoxPlaceholderBackground = 'grey' | 'transparent';
```
ImreBognarUltimaker: We have discussed moving colors into the realm of css, so 'grey' will be changed here to 'light', and wherever else it is needed.

src/components/__tests__/box_placeholder.test.tsx
xli12: Thanks for the suggestion: now it only checks whether the expected class has been included or not..
https://github.com/Ultimaker/react-web-components/pull/246/files#diff-83f623b360f5c357d9712ace54ffc68fR26

PR #250, opened 2019-09-20
Use npm instead of yarn

rvdh:


rvdh:
@ianpaschal Can you please have a look at the linting errors?

ChrisTerBeke:
missing package-json.lock?

package.json
```diff
@@ -16,7 +16,7 @@
         "copyStylesheets": "copyfiles -f ./src/stylesheets/**/* ./dist/stylesheets/",
         "copyFonts": "copyfiles -f ./src/fonts/**/* ./dist/fonts/",
         "cleanDist": "rimraf dist",
-        "prepublishOnly": "yarn cleanDist; tsc; yarn copyImages; yarn copyStylesheets; yarn copyFonts",
+        "prepublishOnly": "npm run-script cleanDist; tsc; npm run-script copyImages; npm run-script copyStylesheets; npm run-script copyFonts",
```
ianpaschal: These can just say `npm run`

.gitlab-ci.yml
```diff
@@ -11,21 +11,21 @@ stages:
...
 
 test:
   stage: test
   script:
-    - yarn lint
-    - yarn test:coverage
-    - yarn build
+    - npm run-script lint
+    - npm run-script test:coverage
+    - npm run-script build
```
ianpaschal: `npm run` works fine.

ianpaschal:
Yeah so this is what I mean by the fact that basically none of our repos were actually adhering to the linting for some reason. No idea how that's been the case for so long undetected but the point is that using NPM here is working fine, it's ESLint that got messed up.

ianpaschal:
FIXXXXXXXXXDIT

ianpaschal:
Fixed it using a little script to convert yarn.lock files into package-lock.json files. This doesn't include the new linting rules yet but it does mean we can get NPM working and then I can work out the linting on top of this.

PR #251, opened 2019-09-23
CS-187 Apply Ultimaker global ESLint config

ianpaschal:
This PR applies the global Ultimaker ESLint config to RWC as well as makes all the required fixes to bring it in line.


ianpaschal:
Replaced by #252

PR #252, opened 2019-09-25
CS-187 apply Ultimaker ESLint config (with NPM)

ianpaschal:
This applies our global Ultimaker ESLint config, now based on the converted-to-npm version.

This replaces #251


src/components/profile_image.tsx
```diff
@@ -15,15 +15,13 @@ export interface ProfileProps {
...
-        }
+            && <Image src={imageURL} size={size} />}
         {!imageURL
             && (
                 <CircleIcon appearance="primary" disabled size={size}>
                     <ProfileIcon />
                 </CircleIcon>
-            )
-        }
+            )}
```
xli12: Just to check: we want `}` to follow the closing tag, but `)}` can be in a separate line...?

ianpaschal: 🤷‍♂ I don't really care one way or another but this is what AirBnB's style guide says to do.

ianpaschal: Oh I know why. It's because they want the closing curly bracket to be next to whatever is the end of the expression. Sometimes that's the tag, but if there's a conditional that might mean there's the parenthesis.

PR #256, opened 2019-10-01
Remove yarn from ignorefiles

rvdh:


ianpaschal:
From the other PR:

> One could argue that this should replace it with npm-debug.log... but I'm _so_ over all of this. I don't want to change that in 7 repos or whatever.

PR #257, opened 2019-10-10
Add SCSS linting

matijs:
- Add the alpha version of `@ultimaker/stylelint-config`
- Ignore files that are not our own code
- Disable rules that are not easily fixable
- Autofix using Stylelint and Prettier
- Fix violations


src/stylesheets/_forms_actions.scss
```diff
@@ -2,8 +2,7 @@
     flex: 0 0 auto;
     padding: 2.4rem 0 0;
     display: flex;
-    flex-wrap: wrap;
-    flex-direction: row-reverse;
+    flex-flow: row-reverse wrap;
```
ianpaschal: I'm curious why here we're using consolidated declarations and in other places we're expending them... I think it was a transition shorthand that got expanded into transition-duration and what not. Also above we use the shorthand for flex.

src/stylesheets/_grid_component.scss
```diff
@@ -14,17 +14,15 @@
         margin: 0;
 
         @include font-size-header-2;
-
-        font-weight: normal;
         font-family: $font-primary;
         font-weight: 500;
-        padding: 0 1.8rem 1.2rem 1.8rem;
+        padding: 0 1.8rem 1.2rem;
```
ianpaschal: Here's another example of short hand being used.

src/stylesheets/_context_menu.scss
```diff
@@ -14,13 +14,16 @@
         user-select: none;
         z-index: 99;
 
-        &:hover, &:focus {
+        &:hover,
+        &:focus {
             background-color: $context-menu-color-hover;
             color: $context-menu-color-trigger-hover;
-            transition: all 0.2s ease-out;
```
ianpaschal: Why is this shorthand removed?

src/stylesheets/_input_fields.scss
```diff
@@ -182,9 +184,8 @@
 
     .react-date-picker__inputGroup {
         @extend %input-style;
-
         color: $color-text;
-        font-weight: 400;
+        font-weight: normal;
```
ianpaschal: I prefer 400 so that we don't end up mixing numbers and string weights.

src/stylesheets/_input_fields.scss
```diff
@@ -212,7 +213,7 @@
             transform: translate(-50%, 0) rotate(45deg);
             top: -6px;
             width: 1rem;
-            content: '';
+            content: "";
```
ianpaschal: Single quotes is more in line with what we did other places.

src/stylesheets/_navigation.scss
```diff
@@ -71,7 +73,7 @@
     }
 
     .slide-in-panel__content {
-        padding: 0 0 1.2rem 0;
```
ianpaschal: I think it's more clear if not writing a single value or double value to specify all 4 sides.

src/stylesheets/_typography.scss
```diff
@@ -5,49 +5,42 @@
...
     font-size: 1.2rem;
 }
 
 h2 {
     font-family: $font-primary;
-    font-weight: 500;
     line-height: 1.24;
     margin: 3.6rem 0 2.4rem;
-    font-weight: normal;
+    font-weight: 500;
```
ianpaschal: See, here we switch from `"normal"` to `500` and in the other place from `400` to `"normal"`. Yuck. Numbers should be used everywhere.

src/stylesheets/_user_account_menu.scss
```diff
@@ -18,14 +18,17 @@
...
 
-        &:hover, &:focus {
+        &:hover,
+        &:focus {
             background-color: $context-menu-color-hover;
             color: $context-menu-color-trigger-hover;
-            transition: all 0.2s ease-out;
+            transition-duration: 0.2s;
+            transition-property: background-color, color;
+            transition-timing-function: ease-out;
```
ianpaschal: Should use shorthand here since we use shorthand for most other props.

src/stylesheets/_variables.scss
```diff
@@ -1,18 +1,18 @@
...
 $color-blue-lightest: #d8f3fd !default;
 $color-red: #ef5b5b !default;
 $color-green: #23aa1d !default;
 $color-orange: #fbb045 !default;
-$color-black: #000000 !default;
-$color-grey-darker: #666666 !default;
-$color-grey-dark: #999999 !default;
+$color-black: #000 !default;
+$color-grey-darker: #666 !default;
+$color-grey-dark: #999 !default;
```
ianpaschal: I prefer 6 digit colors to keep it consistent with all the other colors that are 6 digits.

src/stylesheets/_variables.scss
```diff
@@ -23,13 +23,25 @@ $color-transparent-grey-darkest: rgba(0, 0, 0, 0.85) !default;
 $color-transparent-blue: rgba(0, 167, 233, 0.06) !default;
 
 $font-primary: "Noto Sans", sans-serif !default;
-$font-primary-medium: "Noto Sans", sans-serif !default; // <- We keep this because we reference it in other places but it should never be used ever again! Specify weights instead!
+// We keep the next line because we reference it in other places but it should never be used ever again!
+// Specify weights instead!
```
ianpaschal: Not all weights are used though so you might be specifying an unavailable weight. By using the variables you avoid that. Any reason you don't like the variables?

src/stylesheets/fonts.scss
```diff
@@ -38,55 +38,55 @@
...
 }
 @font-face {
-    font-family: 'Noto Sans';
+    font-family: "Noto Sans";
     font-style: italic;
     font-weight: normal;
-    src: url('../fonts/NotoSans-Italic.woff') format('woff');
-    src: url('../fonts/NotoSans-Italic.woff2') format('woff2');
+    src: url("../fonts/NotoSans-Italic.woff") format("woff");
+    src: url("../fonts/NotoSans-Italic.woff2") format("woff2");
```
ianpaschal: Single quote is more in line with our other style guides. Better for the muscle memory.

src/stylesheets/_context_menu.scss
matijs: Because only `background-color`  and `color` are transitioned, not `all`.

matijs: If you'd like me to, I can rewrite it as two sets of shorthand for `transition`. The way I've solved it here reuses the duration and timing-method instead of duplicating them. I have no preference though, just let me know what you prefer.

src/stylesheets/_input_fields.scss
matijs: I'll update the rule in stylelint-config accordingly.

src/stylesheets/_input_fields.scss
matijs: If those other places are SCSS then that will get updated accordingly…

src/stylesheets/_navigation.scss
matijs: If you feel very strongly about this, we could change the https://stylelint.io/user-guide/rules/shorthand-property-no-redundant-values rule, but I would prefer to keep it this way.

src/stylesheets/_variables.scss
matijs: I prefer the shorter notation. I'll wait for @ImreBognarUltimaker to voice his opinion before changing this.

src/stylesheets/fonts.scss
matijs: That's for not-SCSS, this is SCSS. Right?

matijs: Perfectly reasonable for me, but maybe slightly tenuous for others explanation:

HTML attributes are quoted with double quotes. I'd like to see those double quotes again in selectors like `input[type="password"]` and once you're using them there… mixing them with single quotes is yuck.

ianpaschal: Yes. But we intentionally try to keep things as similar across languages as possible to avoid constantly writing the wrong thing out of muscle memory. SCSS and JS will likely be written by the same people so there's also no reason to arbitrarily change between the languages.

src/stylesheets/_variables.scss
ianpaschal: Yup. My mistake.

src/stylesheets/_user_account_menu.scss
ianpaschal: I see why you did it. My mistake.

src/stylesheets/_input_fields.scss
ianpaschal: They are not. We should be as consistent cross-language when possible. Arbitrarily making a change like this that is going to be at odds with my muscle memory.

src/stylesheets/fonts.scss
matijs: I don't think this is arbitrary. CSS closer to the HTML that it works on than it is to JS. As for the muscle memory argument… in TSX we have to type single quotes, double quotes and backticks.

ianpaschal: Well this was one of the things Imre and Alan and I discussed with the web guys way back and we agreed that we'd use single quotes whenever possible in all front-end code. I can't think of any situation where double quotes are not a violation of the linting rules. And you can't make the "SCSS is closer to HTML" argument; it's a separate language just like JS and HTML.

ImreBognarUltimaker: I agree with Ian on this one.

src/stylesheets/_variables.scss
ImreBognarUltimaker: I like the 6 digit better, in my opinion this is the kind of shorthand that has more overhead than writing out the full code every time.

matijs: I'll have to just find solace in the fact that @ianpaschal prefers `rgb()` notation anyway…

src/stylesheets/fonts.scss
matijs: Changed all double quotes to single quotes. Going to cry in a corner now.

ianpaschal: _hugs_

PR #258, opened 2019-10-15
Fix ESLinting

matijs:
Change the "lint" and "lint:fix" scripts in package.json so ESLint actually lints all it needs to lint.

-  Auto fix linting violations
-  Manually fix linting violations
-  Override linting in search_field.test.tsx to keep test passing, but add a comment that the test needs to be rewritten.


src/components/input_fields/__tests__/search_field.test.tsx
```diff
@@ -50,9 +50,12 @@ describe('The search field component', () => {
         props.staticField = true;
         props.children = 'a child';
         wrapper.setProps(props);
-        props['inputChildren'] = props['children'];
+        // This test needs to be rewritten
```
xli12: Is that possible to elaborate the reason why this test needs to be rewritten?

src/jestsetup.tsx
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) 2018 Ultimaker B.V.
+/* eslint-disable */
```
xli12: `jestsetup.tsx` exists in multiple applications. Just noticed that for some applications `/* eslint-disable */` is not needed, as the dependencies are installed not only for dev, but also for production. Just want to check if there's any (in)consistency in these projects...

matijs: I've looked into this… https://github.com/Ultimaker/stardust-web-components/pull/31/files#diff-e4403a877d80de653400d88d85e4801aR5 seems to be a nicer way of solving this without having to use inline overrides.

src/components/input_fields/__tests__/search_field.test.tsx
matijs: I think it's not a very well written test. The whole file is a bit meh. In the `beforeEach` there's even an assertion…

PR #263, opened 2019-10-31
Refactor Pill component

matijs:
This is a tiny example in which I've "refactored" the Pill component and introduced a few new patterns.

- Use BEM class selectors
- Avoid nesting in Sass
- Use consts for class selectors in the `.tsx` files and declare them at the top of the file
- Remove comments that only take up space but add nothing that is not already clear from the file itself
- Use `React.FC<Props>` instead of the deprecated `React.StatelessComponent<Props>`

### DIFF
--- a/src/components/pill.tsx
+++ b/src/components/pill.tsx
@@ -1,15 +1,19 @@
 import * as React from 'react';
 import classNames from 'classnames';
 
+const PILL_CLASSNAME = 'pill';
+const PILL_CLASSNAME_ACTIVE = 'pill--active';
+
 export interface PillProps {
-    /** Whether the pill should be shown as active */
     active?: boolean
 }
 
-export const Pill: React.StatelessComponent<PillProps> = ({ active, children }) => {
-    const classes = classNames('pill', { active });
+export const Pill: React.FC<PillProps> = ({ active, children }) => {
+    const classes = classNames(PILL_CLASSNAME, {
+        [PILL_CLASSNAME_ACTIVE]: active,
+    });
 
-    return <div className={classes}>{children}</div>;
+    return <span className={classes}>{children}</span>;
 };
 
 Pill.defaultProps = {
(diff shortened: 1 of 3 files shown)


matijs:
This is a tiny example in which I've "refactored" the Pill component and introduced a few new patterns.

- Use BEM class selectors
- Avoid nesting in Sass
- Use consts for class selectors explicit in the `.tsx` files and declare them at the top of the file
- Remove comments that only take up space but add nothing that is not already clear from the file itself
- Use `React.FC<Props>` instead of the deprecated `React.StatelessComponent<Props>`

src/components/pill.tsx
```diff
@@ -1,13 +1,17 @@
...
 export interface PillProps {
-    /** Whether the pill should be shown as active */
     active?: boolean
 }
 
-export const Pill: React.StatelessComponent<PillProps> = ({ active, children }) => {
-    const classes = classNames('pill', { active });
+export const Pill: React.FC<PillProps> = ({ active, children }) => {
+    const classes = classNames(PILL_CLASSNAME, {
+        [PILL_CLASSNAME_ACTIVE]: active,
```
ImreBognarUltimaker: Why is the const in an array here?

src/components/pill.tsx
```diff
@@ -1,15 +1,19 @@
...
+    });
 
-    return <div className={classes}>{children}</div>;
+    return <span className={classes}>{children}</span>;
```
ianpaschal: Why?

ChrisTerBeke: I think because a pill is usually an  inline element instead of block level element? Probably there's CSS now that overrides the div's block display to inline.

ImreBognarUltimaker: If we do this, the children has to be restricted in some form. Otherwise this can produce invalid HTML.

matijs: Like a `div` a `span` does not mean anything on its own, but user agent CSS makes a `div` a block level element and a `span` an inline element. As far as I could see, the way a `.pill` was used only inline so a `span` seems more appropriate.

ianpaschal: Ok. I always thought of them more as floating divs, but I guess I don't really care too strongly one way or another.

PR #264, opened 2019-11-04
Also lint .ts files

matijs:
Only .tsx files were linted.

Rename the 'lint' and 'lint:fix' scripts in package.json to 'lint.ts'
and 'lint:ts:fix' respectively.
Add 'lint' script to package.json that takes care of all linting.
Add 'lint:fix' to package.json to autofix both TypeScript and SCSS.


ianpaschal:
I find the full extensions more readable. It's a super itty bitty nit pick but is there a reason you went for this?

matijs:
Only that I like brace expansion, but full extensions are more readable, will change.

ianpaschal:
I meant with braces but isn't this sort of thing usually written: `'./src/**/*{.ts,.tsx}'`

PR #266, opened 2019-11-06
Fix dependencies

matijs:
Move from depenencies to devDependencies:
- eslint-plugin-react-hooks

Add to devDependencies:
- @types/enzyme-adapter-react-16
- @types/react

In a React + TypeScript project, the ‘@types/react’ package is necessary for type checking to work properly. However, adding it to ‘@ultimaker/react-web-components’ breaks all tests that use the ‘button.tsx’ component because this component allows a `type="link"` on a button that
does not exist for an `<a>` element.

❌ This component will probably have to be refactored into a ‘button.tsx’ component that outputs a HTML `<button>` element and a ‘button-link.tsx’ component that outputs a HTML `<a>` element with a `href` attribute.


ianpaschal:
Looks like problems being caused by installing a newer version of Typescript.

matijs:
@ianpaschal I don't think so. `@types/react` makes jest+ts-jest aware that something is off…

matijs:
Too far behind and this was merely to get a feeling of what would happen if `@types/react` was added as a devDependency. We will still have to work towards that being possible.

PR #268, opened 2019-11-12
CS-466 fix unintended keys closing dropdowns

ImreBognarUltimaker:
This is to fix all dropdowns closing on other than chrome browsers, by pressing any key.

### DIFF
--- a/src/components/__tests__/drop_down_menu_base.test.tsx
+++ b/src/components/__tests__/drop_down_menu_base.test.tsx
@@ -1,10 +1,11 @@
 // Copyright (c) 2018 Ultimaker B.V.
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount, ShallowWrapper } from 'enzyme';
 
 // component
 import { DropDownMenuBase, DropDownMenuBaseProps } from '../drop_down_menu_base';
 import DropDownMenuItem from '../drop_down_menu_item';
+import Button from '../button';
 
 // mocks
 import { mockClickEvent } from '../../__mocks__/clickMock';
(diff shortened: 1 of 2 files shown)


src/components/__tests__/drop_down_menu_base.test.tsx
```diff
@@ -66,7 +67,8 @@ describe('The DropDownMenuBase component', () => {
     });
 
     it('should call toggle menu handler when escape is pressed', () => {
-        wrapper.instance()._onOutsideFocusHandler({ key: 'Escape' });
+        // TODO this needs a better way to test, along with the click event, as right now it is not meaningful.
+        wrapper.instance()._keyPressHandler({ key: 'Escape' });
```
xli12: Have seen a few tests that are actually mocking the user "click" buttons to perform actions. One example:
https://github.com/Ultimaker/stardust-mycloud/blob/staging/src/views/__tests__/ClustersView.test.tsx#L229

Would it be possible to implement something similar, so that we can compare the components that are in the DOM, before and after the updates?

xli12: Perhaps it is meaningful to illustrate why `as right now it is not meaningful`?

ImreBognarUltimaker: This is different, because you have to simulate a click somewhere on the document, outside this component. It is attaching a listener to the document, through which you'd have to trigger a method.
Here it is not meaningful, because a private method is just called on the component. If I add the proper type to the wrapper, this is not even possible.

ImreBognarUltimaker: Added some more details to TODO.

PR #270, opened 2019-11-13
CS-255 Image file size exceeds warning message

xli12:
Issue: `<ImageUploadField />` has a property `maxMb`. Before this PR, when the size of the uploaded image exceeds `maxMb`, it gives no indication for the user what went wrong. Now we add a warning message.

See [CS-255](https://jira.ultimaker.com/browse/CS-255) for more details.

Here is the update for the common component, namely `<ImageUploadField />`.

The preview images below show how the change looks like in the Contributor Portal:
Screenshot 2019-11-13 at 13 57 26

Screenshot 2019-11-13 at 13 57 53

### DIFF
--- a/src/components/image_upload.tsx
+++ b/src/components/image_upload.tsx
@@ -194,7 +194,7 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
         const hoverAreaClasses = classNames('hover-area', { show: dropActive || dropFocus });
 
         return (
-            <div style={{ height: size, width: size }}>
+            <div className="image-upload__dropzone" style={{ height: size, width: size }}>
                 <Dropzone
                     accept="image/jpeg, image/png"
                     multiple={false}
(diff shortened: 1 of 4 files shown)


src/components/image_upload.tsx
```diff
@@ -46,6 +47,10 @@ export interface ImageUploadProps {
     placeholderType?: ImagePlaceholderType;
     /** The maximum amount of megabytes allowed to be uploaded */
     maxMb?: number;
+    /** The warning message to show when the uploaded image file size exceeds maxMb.
+     * Can be any[] if returned from I18n.format.
+    */
+    fileSizeExceedsMessage?: string | any[];
```
matijs: `I18n.format` always outputs a string (according to its type…)

src/components/image_upload.tsx
```diff
@@ -214,12 +228,26 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
         );
     }
 
+    private _renderFileSizeExceedsMessage(): JSX.Element {
```
matijs: Why an extra method? You can just put this in `render()` below.

src/components/image_upload.tsx
```diff
@@ -214,12 +228,26 @@ export class ImageUpload extends React.Component<ImageUploadProps, ImageUploadSt
...
+        const { fileSizeExceedsMessage, size } = this.props;
+        return (
+            <div className="image-upload__warning" style={{ width: size }}>
```
matijs: The class selector `image-upload__warning` seems to be unused. Is the `style` attribute necessary or could this be done using CSS instead?

src/components/input_fields/image_upload_field.tsx
```diff
@@ -36,6 +36,11 @@ export interface ImageUploadFieldProps extends InputFieldProps {
     /** The maximum amount of megabytes allowed to be uploaded */
     maxMb?: number;
 
+    /** The warning message to show when the uploaded image file size exceeds maxMb.
+     * Can be any[] if returned from I18n.format.
+    */
+    fileSizeExceedsMessage?: string | any[];
```
matijs: Same, `any[]` can be removed.

matijs:
As a boy/girl scout rule, we could start removing all of the comments that do not add any value whenever we touch files.

src/components/image_upload.tsx
xli12: The attribute `style` is necessary here for specifying the size, because using `<ImageUploadField />` requires the developer to specify the size of this field. The size varies from one image upload field to another, therefore it cannot be specified in CSS.

src/components/image_upload.tsx
xli12: The `<InputFieldValidation />` components only has a few lines so that is fine to put in `render()`. However I would prefer to put components that have more lines (complex logic/functions etc.) in functions...

PR #272, opened 2019-11-18
CS-415 Minimum amount of code needed to prevent duplicates

ianpaschal:
For now, this just doesn't add tags a 2nd time if they're duplicates. It's the super simple fix to the issue we're having. That being said, maybe we want a better user experience, so that requires some discussion.


sedwards2009:
In the back-end counterpart to this I've made it treat the tags as case insensitive and to basically take the first version of the tag and to drop any duplicate which comes after it.

PR #274, opened 2019-11-21
Remove unneeded prop-types

matijs:
We use TypeScript.


xli12:
Wondering if `prop-types` is used as a dependency of other installed packages?

matijs:
It is a dependency of many other packages but they will install it themselves (check the output of `npm ls prop-types`) this is why it was only removed from package.json and not from package-lock.json.

We don't use it ourselves because we have TS for typing.

PR #275, opened 2019-11-24
CS-447

matijs:
This is the first step to turn react-web-components into a component library with highly re-usable components. 

In its current state, most components are very specific, and nowhere near as re-usable as they ought to be. An `<InputField>` element comes with all sorts of wrappers, validations, labels and even layout!

This PR introduces 2 new icons, a small and normal sized 🔎. The corresponding functional components only output an SVG, as it says on the tin.[1]

It also introduces a highly re-usable new <InputText /> component that is just that.

And finally there's a new simpler <SearchBar /> to replace the old <SearchField />.

The individual commits should cover the thinking behind these new components.

Clean up more (dev)dependencies.

[1] icons will probably move to a shareable repository of their own in due time.


xli12:
I guess that is part of CS-447 (search field) rather than CS-236 (bug fix, remaining time)... right?

src/components/icons/magnifier_icon.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const MagnifierIcon: React.FC<React.SVGAttributes<SVGElement>> = (props) => (
```
xli12: Could you explain why we have two magnifier icons?
I've put these two icons in the browser to see how they look like. The difference between them is that the bigger one has more details. Just wondering the reason why we need two magnifier icons...

src/components/input/__tests__/input_text.spec.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
xli12: Our tests are, at this moment, using the extension `.test.tsx` instead of `.spec.tsx`. Shall we keep using the same extension?

src/components/input/input_text.tsx
```diff
@@ -0,0 +1,40 @@
...
+const INPUT_TEXT_ERROR_CLASSNAME = 'input-text--error';
+
+interface InputTextProps extends React.InputHTMLAttributes<HTMLInputElement> {
+    hasError?: boolean;
+    // prevent creating radio buttons and checkboxes
+    type?: 'text' | 'password' | 'email' | 'url' | 'number';
+}
+
+const InputText = React.forwardRef(
+    (props: InputTextProps, ref: React.Ref<HTMLInputElement>) => {
```
xli12: I think it is fine to keep the possibility to use `ref` here in case the use of `props` is not appropriate; do you have an idea if there's any case in our applications using `ref` and ref-forwarding?

src/components/search_bar.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
```
xli12: (Can be over-engineering but) shall we consider adding more properties to the search bar, for example, text placeholder, which is very common in a search bar?

src/components/icons/magnifier_icon.tsx
matijs: Displaying the more detailed icon at small sizes makes it look ugly so the designers have come up with icons that should be used up to 24x24 pixels and 25x25 and up.

src/components/input/__tests__/input_text.spec.tsx
matijs: Possibly, although we could start using .spec.* instead of .test.* to indicate that a test is a newer test.

src/components/input/input_text.tsx
matijs: I think this is the most React way to create highly reusable components. Instead of dealing with one use case inside a component (for example setting focus after rendering), it's left up to the consuming component to do (_or not do_) something with a `ref` that's forwarded to the child component.

Either way, it's trivial to add them like this for simple components like this that output a DOM element. It would not make sense to add them to more complex components that output a much larger DOM fragment.

src/components/search_bar.tsx
matijs: Placeholder makes sense, I'll add that.

matijs:
@xli12 I think you're probably right about the ticket number, I've changed it accordingly.

src/components/input/__tests__/input_text.spec.tsx
ianpaschal: I prefer .test.tsx.

ianpaschal:
I've slightly changed my opinion on this. A reusable component needs to have exactly the right amount of reusable stuff in it to make it handy but not monolithic. While the RWC components were definitely over engineered, to me, this implementation is too minimalist to be reusable. I mean, it is, but so is the built in HTML `<input/>` tag.

This feels a bit afraid to commit to a specific result because maybe you want it different someday (probably we don't, and if we want to change it we want to change it everywhere). The result is that if you want to reuse this component and expect to get a nice prefab styled component you can basically drag and drop into the UI, you need to go hunting for existing uses somewhere else in our many code bases, check how its used, mimic that, etc. You're essentially re-making the input from scratch and will have to update every single instance in every single repo if you ever want to update all the [insert component here].

Also as I expressed yesterday, layout is to me, one of the most frustrating things in CSS and one which I would most like to have baked into the components. Colors, typefaces, that's easy to just swap in and out. But when we get all the spacing of the icons and text and stuff right on the input, I want it solidified that way and consistent everywhere.

src/components/input/__tests__/input_text.spec.tsx
sedwards2009: What is wrong with just `.test.tsx`?

matijs: Nothing, it's just a name. I'll change it to .test.tsx when reviews are done.

matijs:
Can you please give suggestions on how you would do this? Specifically for `<InputText>` if you can.

Nothing prevents us from having a more elaborate component. In a way the `<SearchBar>` already is a slightly more complex use case of `<InputText>`. I doubt that `<InputText>` would ever be used directly. However, if that were ever needed, it's not possible now because we lack that component unless you actually use `<input>` directly.

As for `<Button>`, which is also a leaf element, that could be a little bit more complex to accommodate the use of icons for example.

xli12:
Since there's a previous update for the version of the dependencies (`package.json`), would it be possible for you to introduce that new update to solve the merge conflict?

matijs:
@xli12 no problem I always rebase first anyway.

ImreBognarUltimaker:
If there is a consensus that this file structure is better and we want to do this, I'm fine with the rest of the changes.

ImreBognarUltimaker:
Actually, If we can make it work, I'd prefer to have all the files for a specific component on the same level. So in the input directory, it would look like this:
```
-input
|-index.ts
|-input_text.tsx
|-input_text.test.tsx
|-input_text.test.tsx.snap
```

This would make it easier to navigate the files and would give a better overview.

matijs:
@ImreBognarUltimaker that has other implications. We use **eslint-config-airbnb-base** which has [this rule](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L71) which, as [line 76](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L76) shows is the de facto standard for Jest. I don't think we should deviate from that. 

Besides, once we start factoring out private functions into utilities or component helpers, the component directory might start to contain other files as well.

ianpaschal:
It's more of a conversation than a specific change request. In my mind, when I have a component that makes my life easier, that means I can just slap in that component into my UI mark-up and it looks and behaves the way I want it.

I guess right now, my gripes are:

1. This feels like a lot of code which ultimately just gives me a plain-old `<Input />` tag. It feels like an entry back into the world of wrappers on top of wrappers so that every component is ✨ _H Y P E R - R E U S A B L E_ ✨, which it just doesn't really need to be.

2. There isn't any default styling tied to it which means every time I use this, I also need to re-write the styles or figure out from existing examples what style selectors to use.

Both of those, particularly the 2nd one, make me question, "Why would I use `<InputText />` vs. `<Input />`.

Going to "Request changes" so that we can continue the conversation without this accidentally getting merged.

src/components/input/input_text.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import classNamesDedupe from 'classnames/dedupe';
+
+const INPUT_TEXT_CLASSNAME = 'input-text';
+const INPUT_TEXT_ERROR_CLASSNAME = 'input-text--error';
+
+interface InputTextProps extends React.InputHTMLAttributes<HTMLInputElement> {
+    hasError?: boolean;
+    // prevent creating radio buttons and checkboxes
+    type?: 'text' | 'password' | 'email' | 'url' | 'number';
```
sedwards2009: Although HTML stuffs number input functionality into the standard INPUT element, inputting a number is somewhat different than normal text. For example, numbers can have min/max constraints, could also be replaced with a slider, and also often have small buttons for increment/decrement it.

Number input should probably be a different component. A password component with a "reveal" button built in, may also make more sense as a separate component.

matijs: If it were ever needed, this low level component would allow you to generate an `<input type="number" min="0" max="100" />`, a slide would be `type="range"` but I don't think we'd need that kind of input anytime soon.

As for password entry with a reveal button or something, that's a component that would use this one.

My issue was that I needed an `<input class="I-don-t-want-to-have-to-think-about-this">` but basically could not have it. @ianpaschal I think this might address at least one of your concerns as well.

matijs:
@ianpaschal to address ✨ H Y P E R - R E U S A B I L I TY ✨… I don't think that all components need to be reusable, just the really low level ones such as these.

ianpaschal:
I figured that's what you were going for, but to me it still raises the question of why whatever level of component is built on top of this isn't just using `<Input />` tags as children.

matijs:
I'd say because just `<input />` (lowercase i) elements do not come with styling that's baked into `<Input />` (capital i). `<Input />` is predictable and maintainable. `<input />` could be anything really.

ianpaschal:
Er yes. Lowercase i. I should have been writing that.

But what styling is baked into `<Input />` then? I'm not seeing it in this PR.

Also it couldn't be anything, really, right? It's a native DOM component, the leafiest leaf.

matijs:
https://github.com/Ultimaker/react-web-components/pull/275/files#diff-5ae90b051fab5a7cb9671c939142f193

And no, it couldn't be anything, but it could look like anything (which could be useful, but not for something that we have said should look like “our” `input[type="text"]`).

ianpaschal:
I saw that but that's not really enough to stand by itself.

So is the answer to my concerns that this is just a building block that other components will use and those other components will also have some specific CSS that means they are "drag and drop" into a UI and I don't have to carefully figure out what combination of CSS classes I need to use?

PR #277, opened 2019-11-28
CS-139 Turn off autocomplete for all number input fields

sedwards2009:
CS-139

This turns autocomplete off for all number input fields.

It is not configurable.


sedwards2009:
What I want is for the browser not to offer completions for number inputs (which are almost always wrong).

As for whether this should be configurable, I don't know but I also can't think of a use case that requires it.

ChrisTerBeke:
@rvdh this PR is not picked up

rvdh:
@ChrisTerBeke started it manually, but it fails.

matijs:
@sedwards2009 any word on the bug?

sedwards2009:
@matijs the jira issue is back in TODO and anyone who can reproduce the issue (I can't) is free to pick it up.

This one-liner PR can be deleted as far as I'm concerned.

matijs:
Deleting this until the bug becomes current again.

PR #278, opened 2019-12-02
Improve new search bar accessibility

matijs:
Add a (hidden) label for the new SearchBar component.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -19,7 +19,7 @@ module.exports = {
                 test: /\.tsx?$/,
                 loaders: [
                     'awesome-typescript-loader',
-                    'react-docgen-typescript-loader',
+                    // 'react-docgen-typescript-loader',
                 ],
                 include: path.resolve(__dirname, '../src'),
                 exclude: /node_modules/,
--- a/src/index.ts
+++ b/src/index.ts
@@ -77,6 +77,12 @@ export * from './components/tooltip';
 export * from './components/user_account_menu';
 export * from './components/video_player';
 
+export * from './components/input';
+export { InputText } from './components/input/input_text';
+export { SearchBar } from './components/search_bar';
+// once icons are in their own package this will be removed
+export * from './components/icons';
+
 // input fields components
 export * from './components/input_fields/checkbox_field';
 export * from './components/input_fields/children_field';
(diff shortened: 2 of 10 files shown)


src/index.ts
```diff
@@ -77,6 +77,12 @@ export * from './components/tooltip';
 export * from './components/user_account_menu';
 export * from './components/video_player';
 
+export * from './components/input';
+export { InputText } from './components/input/input_text';
+export { SearchBar } from './components/search_bar';
+// once icons are in their own package this will be removed
+export * from './components/icons';
```
ianpaschal: Icons are components too; I'd prefer to avoid ninety billion PRs and release and npm installs required when we make changes.

matijs: Good, should anything change, then there'll only be a few so we can deal with that.

xli12:
The content of the PR is fine with me, but there's one thing not very clear: the rationale of this improvement. What will be the use case of this `visually-hidden` style? By introducing a `UUID` field, what kind of issues can we possibly address?

matijs:
@xli12 all input fields need a label but sometimes design does not think that making that label visible is necessary, in those cases we should visually hide it. Using a hardcoded value to link a `<label>` to an `<input>` would break if component appeared more than once on a page, therefore a UUID is used.

src/index.ts
ianpaschal: My point being that I disagree that icons should be a different package.

.storybook/webpack.config.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
                 test: /\.tsx?$/,
                 loaders: [
                     'awesome-typescript-loader',
-                    'react-docgen-typescript-loader',
+                    // 'react-docgen-typescript-loader',
```
matijs: @sedwards2009 found that disabling this speeds up the build no end. Is anybody using the functionality provided by [react-docgen-typescript-loader](https://github.com/strothj/react-docgen-typescript-loader)?

src/stylesheets/_accessibility.scss
```diff
@@ -0,0 +1,10 @@
+@mixin visually-hidden {
+    border: 0;
```
ImreBognarUltimaker: Isn't there a less hacky way to do this?

matijs: This has been the de facto way to visually hide content for quite some time.

src/index.ts
matijs: Fine, we'll discuss that when we get there :) This PR still contains the icons in case you had missed that.

PR #281, opened 2019-12-02
CS-533 GitHub actions

rvdh:
Note: unsure if the publish on tags bit works like this (documentation is a bit lacking).


.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
+name: Build and test
+on:
+  push:
+    branches-ignore:
+      - master
+      - staging
```
ChrisTerBeke: There is no staging branch on this repo.

PR #284, opened 2019-12-10
CS-215 Remove the pop up from the error message component

sedwards2009:
This just removes the pop up from the error message component.

### DIFF
--- a/src/components/response_error.tsx
+++ b/src/components/response_error.tsx
@@ -57,10 +57,10 @@ export interface ResponseErrorProps {
     errorMessageTemplates?: { [key: string]: () => string };
     fieldNames?: { [key: string]: () => string };
 
-    popupHeaderText: string; // Deprecated
-    popupBodyText: string; // Deprecated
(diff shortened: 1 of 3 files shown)


src/components/response_error.tsx
```diff
@@ -58,96 +57,26 @@ export interface ResponseErrorProps {
...
     fieldNames?: { [key: string]: () => string };
 
-    popupHeaderText: string;
-    popupBodyText: string;
-    popupDownloadButtonText: string;
-    popupCancelButtonText: string;
+    popupHeaderText: string; // Deprecated
+    popupBodyText: string; // Deprecated
+    popupDownloadButtonText: string; // Deprecated
+    popupCancelButtonText: string; // Deprecated
```
ianpaschal: We should remove them if they're not used. Otherwise your IDE will suggest things you need but actually don't. Even optional property would be an option. But removing is best.

sedwards2009: I'm hesitant to be forced into updating a whole bunch of related front-end repos just for this.

PR #289, opened 2019-12-23
CS-498 Replace StatelessComponent with FC

xli12:
`React.StatelessComponent` has been renamed to `React.FunctionComponent` or `React.FC`.

https://twitter.com/dan_abramov/status/1057625147216220162


matijs:
@rvdh there seems to be an issue with publish here :]

rvdh:
@matijs fixed!

PR #293, opened 2020-01-15
Remove the border from the ImageCropper

matijs:
We use avatar images at 180x180 pixels on the profile page of
account.ultimaker.com but we only provide a 130x130px circle when cropping
images. By removing the 25px border that was previously used, we get the
full 180x180px of the cropper tool.

**before**
image
**after**
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@ultimaker/react-web-components",
-    "version": "5.9.2",
+    "version": "5.9.3",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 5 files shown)


ImreBognarUltimaker:
Any idea why was it there in the first place?

ChrisTerBeke:
Looks a lot better, I'd say go for it.

matijs:
@ImreBognarUltimaker nope, it happened in [this commit](https://github.com/Ultimaker/react-web-components/commit/79b679d5a129649f3a177c5dd842fc8db5d624a5) with a crappy commit message so it's lost forever.

PR #294, opened 2020-01-21
Improve SearchBar

matijs:
A SearchBar is responsible for its own internal state and rendering once
it's mounted.

This change allows a parent component to force SearchBar to update its
internal state (the value of query) to a new value.

Geneate a uuid only once.

### DIFF
--- a/src/components/search_bar/search_bar.tsx
+++ b/src/components/search_bar/search_bar.tsx
@@ -21,6 +21,8 @@ class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
         query: '',
     }
 
+    id = uuidv4();
+
     debouncedEmit = (({ onChange, emitDelay }) => debounce(
         onChange, emitDelay,
     ))(this.props);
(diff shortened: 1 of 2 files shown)


src/components/search_bar/search_bar.tsx
```diff
@@ -40,25 +43,33 @@ class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
         this.debouncedEmit.cancel();
     }
 
     onChange(event: React.ChangeEvent<HTMLInputElement>) {
         const { value: query } = event.target;
-
         this.setState({
             query,
+        }, () => {
+            this.debouncedEmit(query);
```
ImreBognarUltimaker: Ha!

src/components/search_bar/search_bar.tsx
```diff
@@ -40,25 +43,33 @@ class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
+        const { query } = this.props;
+        if (query !== prevProps.query && query !== prevState.query) {
+            // eslint-disable-next-line react/no-did-update-set-state
+            this.setState({
+                query,
+            });
+        }
+    }
+
     componentWillUnmount() {
```
xli12: Would it be fine to replace it with `componentDidUpdate()`?
https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate

matijs: This is `componentWillUnmount`, not `componentWillUpdate`

src/components/search_bar/search_bar.tsx
matijs: In practice this made no difference but it's still nicer to make sure that the debounced handler is called after state was updated.

PR #296, opened 2020-02-04
CS-695: Fix package vulnerabilities

matijs:
Running `npm audit fix` automatically fixes all but one of the vulnerabilities.

The only vulnerability left is a [moderate one](https://npmjs.com/advisories/812) in `@storybook/addon-info` that is only used in development.

All packages that could be updated without tests or the build failing
were updated to the last versions.

Updating eslint-plugin-react required a few fixes where code was
violating the linting rules.

Updating to the latest version of Storybook is not trivial. Currently
@dump247/storybook-state is used, but this package does not work with
the next major version of Storybook. This, in addition to some typing
issues, is the biggest hurdle to upgrade to the latest version of
Storybook.


ImreBognarUltimaker:
Works on my machine.

PR #327, opened 2020-02-06
Remove unused package, add missing types

matijs:
Add missing types for `node-gettext`.
Remove `gettext-extractor` as it's no longer used. Dependabot should automatically close its open PR for this as well.


package.json
```diff
@@ -69,6 +69,7 @@
         "@types/jest": "^24.9.1",
         "@types/lodash.debounce": "^4.0.6",
         "@types/node": "^12.12.26",
+        "@types/node-gettext": "^2.0.1",
```
ChrisTerBeke: I think node-gettext and gettext-parser are also not needed anymore (they are for the same functionality as gettext-extractor was used for).

matijs: i18n.tsx uses it…

matijs: That also uses gettext-parser by the way…

ChrisTerBeke: ah right, we should leave those in then (and only remove gettext-extractor).

PR #338, opened 2020-02-11
CS-781 Implement new UM blue

ImreBognarUltimaker:
There were no weird stuff in the storybook after the change.

### DIFF
--- a/src/stylesheets/_variables.scss
+++ b/src/stylesheets/_variables.scss
@@ -139,7 +139,6 @@ $context-menu-disabled-color-trigger: $color-grey-light !default;
 $context-menu-disabled-color-trigger-hover: $color-background-dark !default;
 $context-menu-disabled-color-hover: $color-transparent !default;
 
-$checkbox-color-border-active: $um-blue-5 !default;
 $checkbox-color-tick-active: $color-white !default;
 $checkbox-color-border: $secondary-btn-color-border !default;
 $checkbox-color-tick-hover: $color-border !default;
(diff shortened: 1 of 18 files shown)


ChrisTerBeke:
Do we also need to update the button/link hover color (and similar stuff)?

ImreBognarUltimaker:
> Do we also need to update the button/link hover color (and similar stuff)?

Yes, part of the reason we do this is to expose what more needs to be done exactly.

src/stylesheets/_variables.scss
```diff
@@ -61,16 +95,16 @@ $color-background-success: $color-green !default;
 
 $color-border: $color-grey-light !default;
 $color-border-light: $color-grey-lighter !default;
-$color-border-active: $color-blue !default;
+$color-border-active: $um-blue-5 !default;
```
matijs: We want to move away from this. These variables that refer another `$color-*` variable are used throughout the code base but we need to selectively go through the code base and replace these with the `$um-*` variables where applicable.

I would start by making the button have the same look and feel as the [button](/Ultimaker/stardust-web/blob/master/src/components/buttons/_um-button.scss) in `stardust-web`. Other components can follow in separate smaller pull requests. Coordinate with design to make sure everything looks the way it should.

ImreBognarUltimaker: I think it would be best if you added what needs to be done to the ticket:
https://jira.ultimaker.com/browse/CS-781

It's not mentioned that we want to "refarctor" the CSS as part of this story.

matijs: It was discussed (and probably cut a bit short) during refinement. I'll add it.

ImreBognarUltimaker: Thanks!

PR #347, opened 2020-02-17
Release v7.0.0

matijs:


rvdh:
Why the major version changes? Are these all "incompatible API changes"?

matijs:
@rvdh yes… removal of variables breaks stuff in other repos (in part that's because of how this was set up).

PR #407, opened 2020-04-23
Update workflows

rvdh:


rvdh:
Previously builds succeeded because we used `npm ci`. Now that we're using the generic workflow files, which use `npm install`, some dependencies are being updated and are causing errors.. @daansmit is looking into this.

PR #408, opened 2020-04-23
CS-860 Keep menus inside the main area, and under the header

sedwards2009:
CS-860


sedwards2009:
This approach will cause trouble for modal dialogs.

PR #499, opened 2021-01-25
Create dependabot.yml

ChrisTerBeke:


xli12:
Hmmm... do we still want to enable dependabot for this expected-to-be deprecated repo?

PR #500, opened 2021-03-16
CS-1848 Add copyright notice to `README.md`

sedwards2009:
CS-1848

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+Copyright © Ultimaker 2018-2021 
+
 # React Web Components
 
 **Warning: this library is considered deprecated.


package.json
```diff
@@ -32,7 +32,6 @@
         "url": "git+https://github.com/Ultimaker/react-web-components.git"
     },
     "author": "Ultimaker",
-    "license": "ISC",
```
ChrisTerBeke: this repo is (deprectated but) open source, so we'll need to keep a license agreement in place.

sedwards2009: We might just have to leave this as-is, even though we don't have a copy of the license or any other mention elsewhere, just these three letters.

I don't think the license question was properly asked when this repo was set up either.

sedwards2009: Also, personally I would prefer to use MIT here, given that it is far more common while being functionally the same.

ianpaschal: ^Agree

ianpaschal (approved):

sedwards2009:
This repo is obsolete anyway and having build troubles. closing

